PR #1, opened 2017-10-16
Fixes #11608: Port external node info to rpkg format

fanf:
https://www.rudder-project.org/redmine/issues/11608


fanf:
PR rebased

PR #3, opened 2017-10-25
Fixes #11661: Bad repo declaration in node-external-reports pom.xml

fanf:
https://www.rudder-project.org/redmine/issues/11661


fanf:
Self merging because @VinceMacBuche is trying to do ops things and it is trivial.

PR #25, opened 2018-03-21
Fixes #12255: Datasource compilation error with API provider trait

fanf:
https://www.rudder-project.org/redmine/issues/12255


fanf:
Also need that one for CI integration tests

PR #26, opened 2018-03-21
Fixes #12257: maven furious. maven requires more life sacrifices to be happy

fanf:
https://www.rudder-project.org/redmine/issues/12257


fanf:
self merging to try to do the CI integration

PR #27, opened 2018-03-26
Fixes #12275: "\" in build.conf description leads to unreacoverable exception during boot

fanf:
https://www.rudder-project.org/redmine/issues/12275


fanf:
OK, merging this PR

PR #29, opened 2018-04-11
Fixes #12416: Bad internal API URL in datasources for Rudder 4.3

fanf:
https://www.rudder-project.org/redmine/issues/12416


datasources/src/main/scala/com/normation/plugins/datasources/DataSourcesPluginDef.scala
```diff
@@ -61,7 +61,6 @@ class DataSourcesPluginDef(override val status: PluginStatus) extends DefaultPlu
   override val basePackage = "com.normation.plugins.datasources"
 
   def init = {
-    PluginLogger.info(s"loading '${buildConf.getString("plugin-id")}:${version.toString}' plugin")
```
fanf: This is done by default, no need to keep it (and plugin-id doesn't exists anymore)

PR #35, opened 2018-05-17
Fixes #12646: Make branding plugin ready for classic plugin setup

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12646


pom-template.xml
```diff
@@ -12,7 +12,7 @@
     <groupId>com.normation.rudder</groupId>
     <artifactId>rudder-parent</artifactId>
     <!-- here, we can't use ${rudder-build-version} because of maven limitation-->
-    <version>4.3.0~rc4-SNAPSHOT</version>
+    <version>4.3.2-SNAPSHOT</version>
```
fanf: I prefer that we use 4.3.0 to be sure to be source compatible with that version. We will update that value only in case of a module API breaking change, but it must be something exceptionnal and well documented.

PR #43, opened 2018-06-24
Fixes #12203: Interface to manage ACL and user token for API authorization

fanf:
https://www.rudder-project.org/redmine/issues/12203


fanf:
PR rebased

api-authorizations/pom-template.xml
```diff
@@ -27,9 +30,71 @@
...
+            <phase>process-resources</phase>
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>target/classes/template</outputDirectory>
+              <resources>
+                <resource>
+                  <directory>src/main/elm</directory>
+                  <includes><include>reportingManagement.html</include></includes>
```
VinceMacBuche: I think this file is missing and it should be renamed according to correct plugin name

VinceMacBuche: In fact we do not need that section about the html template we will include an plugin on api account page

api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/snippet/ServeResource.scala
```diff
@@ -35,7 +35,7 @@
 *************************************************************************************
```
VinceMacBuche: ServeResource.scala should be delete

VinceMacBuche:
This needs to be rebased

fanf:
PR rebased

fanf:
OK, merging this PR

PR #44, opened 2018-06-25
Fixes #12834: We need to explicit version of scala-library in plugin-parent else one is pick at random from transitive dependencies

fanf:
https://www.rudder-project.org/redmine/issues/12834


fanf:
OK, merging this PR

PR #47, opened 2018-06-28
Fixes #12848: Duplicate definition of scala-version in parent ticket

ncharles:
https://www.rudder-project.org/redmine/issues/12848


ncharles:
OK, merging this PR

PR #48, opened 2018-06-28
Fixes #12850: Correct packaging for branding plugin

ncharles:
https://www.rudder-project.org/redmine/issues/12850


fanf:
OK, merging this PR

PR #62, opened 2018-07-09
Fixes #12918: Datasource plugin does not declare its APIs

fanf:
https://www.rudder-project.org/redmine/issues/12918


VinceMacBuche:
OK, merging this PR

PR #68, opened 2018-07-16
Fixes #12971: A part of the login form disappears if the custom bar is not activated

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/12971


fanf:
OK, merging this PR

PR #74, opened 2018-07-19
Fixes #13010: Bad image links in api-authorizations doc

fanf:
https://www.rudder-project.org/redmine/issues/13010


fanf:
PR rebased

PR #78, opened 2018-07-23
Fixes #13024: Create alternative backend authentication plugin

fanf:
https://www.rudder-project.org/redmine/issues/13024


fanf:
PR rebased

PR #88, opened 2018-08-02
Fixes #13142: Cronify the execution of the centreon plugin

VictorQrt:
https://www.rudder-project.org/redmine/issues/13142

### DIFF
--- a/centreon/packaging/postinst
+++ b/centreon/packaging/postinst
@@ -1,3 +1,4 @@
 #!/bin/sh
 
 mkdir -p /var/rudder/plugin-resources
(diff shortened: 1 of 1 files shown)


centreon/packaging/postinst
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 
 mkdir -p /var/rudder/plugin-resources
+echo "* */1 * * * /opt/rudder/bin/centreon-plugin.py commit" > /etc/cron.d/centreon-rudder
```
peckpeck: * means every minute
*/1 is equivalent to * and means every hour

VictorQrt:
Commit modified

PR #99, opened 2018-09-05
Fixes #12194: Target groups for the Validation Workflow

fanf:
https://www.rudder-project.org/redmine/issues/12194

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
@@ -43,14 +43,24 @@ import net.liftweb.common._
 import com.normation.rudder.repository._
 import com.normation.rudder.services.eventlog.WorkflowEventLogService
 import com.normation.rudder.batch.AsyncWorkflowInfo
-import com.normation.rudder.services.marshalling.ChangeRequestChangesSerialisation
+import com.normation.rudder.domain.nodes.NodeGroupId
+import com.normation.rudder.domain.parameters.ParameterName
+import com.normation.rudder.domain.policies.DirectiveId
+import com.normation.rudder.domain.policies.GroupTarget
+import com.normation.rudder.domain.policies.Rule
+import com.normation.rudder.domain.policies.RuleId
+import com.normation.rudder.domain.policies.RuleTarget
+import com.normation.rudder.domain.policies.SimpleTarget
+import com.normation.rudder.domain.policies.TargetExclusion
+import com.normation.rudder.domain.policies.TargetIntersection
+import com.normation.rudder.domain.policies.TargetUnion
 import com.normation.rudder.services.workflows.CommitAndDeployChangeRequestService
 import com.normation.rudder.services.workflows.NoWorkflowAction
+import com.normation.rudder.services.workflows.RuleChangeRequest
 import com.normation.rudder.services.workflows.WorkflowAction
 import com.normation.rudder.services.workflows.WorkflowService
 import com.normation.rudder.services.workflows.WorkflowUpdate
 
-import scala.xml.NodeSeq
 
 /**
  * A proxy workflow service based on a runtime choice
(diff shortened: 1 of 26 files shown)


fanf:
PR rebased

change-validation/src/main/scala/com/normation/plugins/changevalidation/snippet/ChangeRequestDetails.scala
```diff
@@ -0,0 +1,377 @@
...
+        } }
+        </h4>
+     </div>
+    }
+    def content(default:(WorkflowNodeId,stepChangeFunction)) = {
+      val classForButton = action match {
+      case "Decline" => "btn-danger"
+      case _ => "btn-success"
+    }
+      ( "#header"   #>  s"${action} CR #${cr.id}: ${cr.info.name}" &
```
VinceMacBuche: id.value instead of id

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
```diff
@@ -63,58 +71,126 @@ class EitherWorkflowService(cond: () => Box[Boolean], whenTrue: WorkflowService,
...
+  }
+
+  /*
+   * Just check if the node group is the list of monitored groups
+   */
+  override def forNodeGroup(actor: EventActor, change: NodeGroupChangeRequest): Box[Boolean] = {
+    for {
+      monitored <- monitoredTargets()
+    } yield {
+      monitored.contains(GroupTarget(change.newGroup.id))
```
VinceMacBuche: What if the targetted groups is based on nodes from a targetted group ?

VinceMacBuche: Let's say you have two groups, A and B with B having criteria (Nodes from group A), and A is a monitored Group.

Do we want to trigger a change request if B is modified ?

VinceMacBuche: I'm also wondering if we should check modifications if the change impact a Node from one of the monitored group, even if the group itself is not concerned

VinceMacBuche: Hence if the correct semantic is 'That group and only that group" it's ok

fanf: The logic is to make any change on targets lead to a request. But nothing here deals with group of groups, because it would need to look at group request and extract any target info (which is not done). 

I think we can still release a first version without that, and I'm opening a ticket to track the bug.

fanf: What change to a node can trigger a change request?

fanf: (I mean, of course inventory/parameter modification can change the fact that a node config change, but we clearly don't track inventory change, and I don't thing we have change requests for parameter?)

VinceMacBuche: Let's say I have a Production group that is monitored with Nodes (N1 and N42) and a Group "Earth champions" not monitored with Node "N42"  in it

I have a Rule Targetting "Earth champions" Group would you expect modification on this group to be applied via a change request or directly ? Since it applies to a Node from the "Production" group ?

fanf: Hum, good question. 
I (personnaly) would not expect it, because I wrote the code, but from a user point of view, I believe it makes very little sense. Worse, I expect people will make "control" group only for change requests (with no application whatsoever), which would break here. 

So, we need to have somehow the list of groups a node belongs to, and intersect that with the list of supervised groups. I propose to add that in the afformentionned ticket :)

fanf:
For reference, the ticket where the proposed update are tracked is here: https://www.rudder-project.org/redmine/issues/13559

PR #100, opened 2018-09-11
Fixes #13442: Branding plugin doesn't build in branch 5.0

ncharles:
https://www.rudder-project.org/redmine/issues/13442

### DIFF
--- a/main-build.conf
+++ b/main-build.conf
@@ -21,4 +21,4 @@ lib-common-private=1.1
 
 # Version of Rudder used to build the plugin.
 # It defined the API/ABI used and it is important for binary compatibility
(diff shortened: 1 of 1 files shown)


main-build.conf
```diff
@@ -21,4 +21,4 @@ lib-common-private=1.1
 
 # Version of Rudder used to build the plugin.
 # It defined the API/ABI used and it is important for binary compatibility
-rudder-build-version=5.0.0~beta2-SNAPSHOT
+rudder-build-version=5.0.0~beta3-SNAPSHOT
```
VinceMacBuche: In fact we should use rc1-SNAPSHOT, or beta2 without snapshot

ncharles:
Commit modified

PR #112, opened 2018-09-26
Fixes #13579: Run centreon-plugin.py commit when we detect a csv changes instead of each hour

peckpeck:
https://www.rudder-project.org/redmine/issues/13579


peckpeck:
Commit modified

peckpeck:
OK, merging this PR

PR #115, opened 2018-09-27
Fixes #13558: Add an UI to display configured authentication backend 

fanf:
https://www.rudder-project.org/redmine/issues/13558


fanf:
PR rebased

PR #119, opened 2018-10-03
Fixes #13559: Check node, not group, for need of change request

fanf:
https://www.rudder-project.org/redmine/issues/13559

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
@@ -1,0 +1,156 @@
+package com.normation.plugins.changevalidation
+
+import com.normation.eventlog.EventActor
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.nodes.NodeInfo
+import com.normation.rudder.domain.policies.Rule
+import com.normation.rudder.domain.policies.SimpleTarget
+import com.normation.rudder.repository.FullNodeGroupCategory
+import com.normation.rudder.repository.RoChangeRequestRepository
+import com.normation.rudder.repository.RoNodeGroupRepository
+import com.normation.rudder.repository.RoRuleRepository
+import com.normation.rudder.services.nodes.NodeInfoService
+import com.normation.rudder.services.workflows.DirectiveChangeRequest
+import com.normation.rudder.services.workflows.GlobalParamChangeRequest
+import com.normation.rudder.services.workflows.NodeGroupChangeRequest
+import com.normation.rudder.services.workflows.RuleChangeRequest
+import net.liftweb.common.Box
+import net.liftweb.common.Full
+
+/**
+ * Check is an external validation is needed for the change, given some
+ * arbitrary rules defined in implementation.
+ */
+trait ValidationNeeded {
+  def forRule       (actor: EventActor, change: RuleChangeRequest       ): Box[Boolean]
+  def forDirective  (actor: EventActor, change: DirectiveChangeRequest  ): Box[Boolean]
+  def forNodeGroup  (actor: EventActor, change: NodeGroupChangeRequest  ): Box[Boolean]
+  def forGlobalParam(actor: EventActor, change: GlobalParamChangeRequest): Box[Boolean]
+}
+
+
+/*
+ * A version of the "validationNeeded" plugin which bases its oracle on a list
+ * of group. The list of group is used to mark nodes.
+ *
+ * Any modification that impacts a marked node triggers a validation:
+ * - a modification in a GlobalParam is always validated (because can be used in CFEngine code)
+ * - a modification in a node group is validated only if it's one of the supervised one
+ * - a modification in a rule is validated only if:
+ *    - one of the modification add or remove a monitored group
+ *    - intersection of the set of nodes targeted by the rule with the set of marked nodes is non empty
+ * - a modification in a directive is validated if it as at least configured in one rule where modification
+ *   are supervised.
+ *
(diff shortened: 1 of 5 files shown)


fanf:
PR rebased

change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
```diff
@@ -0,0 +1,137 @@
...
+    ChangeValidationLogger.Metrics.debug(s"Check rule '${change.newRule.name}' [${change.newRule.id.value}] change requestion need for validation in ${System.currentTimeMillis() - start}ms")
+    res
+  }
+
+  /**
+   * This method checks if at least one of the nodes belonging to rule targets for the change
+   * is supervised.
+   */
+  def checkNodeTargetByRule(groups: FullNodeGroupCategory, allNodeInfo: Map[NodeId, NodeInfo], monitored: Set[SimpleTarget], rules: Set[Rule]): Boolean = {
+    val monitoredNodes = groups.getNodeIds(monitored.map(identity), allNodeInfo)
```
VinceMacBuche: Why do you need to '.map(identity)' ?

fanf: because set is monovariant, and I have a Set[SimpleTarget] where I need a Set[RuleTarget].

change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
```diff
@@ -0,0 +1,137 @@
...
+      checkNodeTargetByRule(groups, nodeInfo, monitored, rules.toSet)
+    }
+    ChangeValidationLogger.Metrics.debug(s"Check directive '${change.newDirective.name}' [${change.newDirective.id.value}] change requestion need for validation in ${System.currentTimeMillis() - start}ms")
+    res
+  }
+
+  /*
+   * Just check if the node group is the list of monitored groups
+   */
+  override def forNodeGroup(actor: EventActor, change: NodeGroupChangeRequest): Box[Boolean] = {
```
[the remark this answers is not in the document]
fanf: you are right

fanf:
PR rebased

change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
```diff
@@ -0,0 +1,156 @@
...
+
+    val start = System.currentTimeMillis()
+
+    val res = for {
+      groups      <- groupLib.getFullGroupLibrary()
+      allNodeInfo <- nodeInfoService.getAll()
+      monitored   <- monitoredTargets()
+    } yield {
+      val targetNodes = change.newGroup.serverList ++ change.previousGroup.map(_.serverList).getOrElse(Set())
+      val exists = groups.getNodeIds(monitored.map(identity), allNodeInfo).find(nodeId => targetNodes.contains(nodeId))
```
VinceMacBuche: Don't we want to use intersect and list all Nodes making a conflict ? could be a problem if there is a lot of Nodes though ...

change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
```diff
@@ -0,0 +1,156 @@
...
+   * is supervised.
+   */
+  def checkNodeTargetByRule(groups: FullNodeGroupCategory, allNodeInfo: Map[NodeId, NodeInfo], monitored: Set[SimpleTarget], rules: Set[Rule]): Boolean = {
+    val monitoredNodes = groups.getNodeIds(monitored.map(identity), allNodeInfo)
+    val changes = rules.flatMap(_.targets)
+    val exists = groups.getNodeIds(changes, allNodeInfo).exists(nodeId => monitoredNodes.contains(nodeId))
+    // we want to let the log knows why the change request need validation
+    if(exists && ChangeValidationLogger.isDebugEnabled) {
+      rules.foreach { rule =>
+        groups.getNodeIds(rule.targets, allNodeInfo).find(nodeId => monitoredNodes.contains(nodeId)).foreach { node =>
```
VinceMacBuche: You could use intersect between the two Sets, and you could also put  groups.getNodeIds(changes, allNodeInfo) in a value before

change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
fanf: I prefer to just give an example, else we are going to list the union of all nodes in faulty group, so perhaps thousand of them. 
And I use the idea that either the user missed a special case (and there will be few nodes on it, so just an example should be ok), or he forgot a whole group (and in that case, just an example of the node should be ok). 

I could also limit to for ex 10 nodes, so that in the case of special cases, it's likely that they are all displayed.

change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
fanf: The find is necessary for to avoid unecessary walk of the set. OK for the variable def.

change-validation/src/main/scala/com/normation/plugins/changevalidation/ValidationNeeded.scala
VinceMacBuche: Ok !! I merge this !! let's try that because and we'll see if it's ok or not when using it

PR #130, opened 2018-10-25
Fixes #13726: Add missing plugin docs: centreon

amousset:
https://www.rudder-project.org/redmine/issues/13726


amousset:
Commit modified

PR #139, opened 2018-11-22
Fixes #13854: Make a feature to open ticket based on rudder events

Fdall:
https://issues.rudder.io/issues/13854


Fdall:
Commit modified

Fdall:
Commit modified

Fdall:
Commit modified

Fdall:
notify/bin/rudder-notifyd is fully commented because I dont think we need it anymore,
thought, I havn't fully tested the functionnalities of the plugin and it may have a sense to fork via python and not use a systemd daemon?

notify/Makefile
```diff
@@ -3,7 +3,8 @@
 # make licensed : will build a license limited version of the plugin
 #
 
-FILES = bin/ etc/ share/
+#FILES = bin/ etc/ share/
```
amousset: why?

notify/bin/rudder-notifyd
```diff
@@ -1,112 +1,114 @@
...
-        elif sys.argv[1] == "restart":
-            d.restart()
-        else:
-            print(usage)
-            sys.exit(2)
-        sys.exit(0)
-    else:
-        print(usage)
-        sys.exit(2)
+# Do not use anymore, deprecated
```
amousset: You should remove it.

notify/packaging/postinst
```diff
@@ -6,14 +6,51 @@ CONFFILE="/opt/rudder/etc/notify.conf"
...
+[GLPI]
+on = false
+userToken = <user token>
+apiToken = <api token>
+url = <API url>
+
+EOF
+fi
+
+SYSTEMDFILE="/etc/systemd/system/notify.service"
```
amousset: the service name should be `rudder-notify`.

notify/share/python/glpi.py
```diff
@@ -0,0 +1,147 @@
...
+    self.__dict__ = fields
+
+  def __init__(self, name, content, status=1, urgency=3, impact=3, priority=3):
+    self.name = name
+    self.content = content
+    self.status = status
+    self.urgency = urgency
+    self.impact = impact
+    self.priority = priority
+#    self.entities_id = entities_id
```
amousset: why are the commented.

Fdall:
PR rebased

notify/share/python/glpi.py
Fdall: They are not used here, but may be in future updates

notify/Makefile
Fdall: There are no files to put in the etc/ folder, which will result in a failure at compilation time.
Some files comes directly from a postinst execution and not from a copy while installing the plugin.

Fdall:
OK, merging this PR

Fdall:
OK, merging this PR

PR #149, opened 2019-01-30
Fixes #14260: change dependency of plugins to Rudder 5.0.5

ncharles:
https://issues.rudder.io/issues/14260


ncharles:
was implemented in https://github.com/Normation/rudder-plugins/pull/152/files

PR #150, opened 2019-02-08
Fixes #7876: [API] Create function to "pre-create" a node

fanf:
https://issues.rudder.io/issues/7876

### DIFF
--- a/create-node-api/Makefile
+++ b/create-node-api/Makefile
@@ -1,0 +1,9 @@
+#
+# make all : will build node-external-reports.rpkg.
+# make licensed : will build a license limited version of the plugin
+#
+
+FILES = # add the file that need to be package here like $(NAME)/create-node-api-schema.sql
+SCRIPTS = postinst
+
+include ../makefiles/common-scala-plugin.mk
--- a/create-node-api/packaging/postinst
+++ b/create-node-api/packaging/postinst
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# add post-installation script, like:
+
+# psql -U rudder -h localhost -d rudder -f /opt/rudder/share/plugins/createnodeapi/Createnodeapi-schema.sql
+
--- a/create-node-api/src/main/scala/com/normation/plugins/createnodeapi/CreateNodeApi.scala
+++ b/create-node-api/src/main/scala/com/normation/plugins/createnodeapi/CreateNodeApi.scala
@@ -1,0 +1,263 @@
+/*
+*************************************************************************************
+* Copyright 2019 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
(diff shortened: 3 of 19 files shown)


create-node-api/Makefile
```diff
@@ -0,0 +1,13 @@
...
+# make all : will build node-external-reports.rpkg.
+# make licensed : will build a license limited version of the plugin
+#
+
+FILES = # add the file that need to be package here like $(NAME)/create-node-api-schema.sql
+SCRIPTS = postinst
+
+include ../makefiles/common-scala-plugin.mk
+
+target/$(NAME)/create-node-api-schema.sql:
```
VinceMacBuche: I think you can remove this

create-node-api/packaging/postinst
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# add post-installation script, like:
+
+# psql -U rudder -h localhost -d rudder -f /opt/rudder/share/plugins/createnodeapi/Createnodeapi-schema.sql
```
VinceMacBuche: Should we keep this comment ? I think we need the file

create-node-api/src/main/scala/com/normation/plugins/createnodeapi/CreateNodeApiPluginDef.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  def init = {}
+
+  def oneTimeInit : Unit = {}
+
+  val configFiles = Seq()
+
+  override def apis: Option[LiftApiModuleProvider[_ <: EndpointSchema]] = {
+    if(status.isEnabled()) {
+      println("******** added apu")
```
VinceMacBuche: I think you can remove this ;) and the oh no println below !!

create-node-api/src/main/scala/com/normation/plugins/createnodeapi/CreateNodeApi.scala
```diff
@@ -0,0 +1,261 @@
...
+        inventory.node.main.id
+      }) match {
+        case Full(id)     => Right(id)
+        case eb: EmptyBox => Left(CreationError.OnSaveInventory((eb?~!"Error during node creation").messageChain))
+      }
+    }
+
+    def accept(template: NodeTemplate): Either[CreationError, NodeSetup] = {
+      val id = template.inventory.node.main.id
+      newNodeManager.accept(id, ModificationId(uuidGen.newUuid), Rest.CreateNodeActor) match {
```
VinceMacBuche: Don't we want to pass the actor calling the api ? identified through it's api token ?

fanf:
ping @fanf

create-node-api/packaging/postinst
fanf: I prefer to keep it.

create-node-api/src/main/scala/com/normation/plugins/createnodeapi/CreateNodeApi.scala
fanf: yes, certainly so

fanf:
Commit modified

create-node-api/Makefile
```diff
@@ -7,7 +7,4 @@ FILES = # add the file that need to be package here like $(NAME)/create-node-api
 SCRIPTS = postinst
 
 include ../makefiles/common-scala-plugin.mk
-
-target/$(NAME)/create-node-api-schema.sql:
-	cp ./src/main/resources/create-node-api-schema.sql target/$(NAME)/
-
+println
```
VinceMacBuche: what does println do ? you need at least a target/command so the makefile is valid  ?

VinceMacBuche: Apart that comment it can be merged !!

fanf:
Commit modified

PR #161, opened 2019-04-04
Fixes #14540: branding doesn't build with elm 0.19

VinceMacBuche:
https://issues.rudder.io/issues/14540


fanf:
It seems that the alpha chan get mixed up with the others. Try to: change colors, save (correct color chosen), logout, relog => bad colors

VinceMacBuche:
Commit modified

PR #168, opened 2019-05-06
Fixes #14826: Select a group of user concerned by the workflow

ElaadF:
https://issues.rudder.io/issues/14826


ElaadF:
Commit modified

ElaadF:
Commit modified

PR #171, opened 2019-06-20
Fixes #15113: Missing logs in change validation plugin

ncharles:
https://issues.rudder.io/issues/15113


change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJdbcRepository.scala
```diff
@@ -189,7 +189,10 @@ class WoChangeRequestJdbcRepository(
     for {
       id  <- (q.transact(xa).attempt.unsafeRunSync: Box[Int])
       cr  <- roRepo.get(ChangeRequestId(id)).flatMap {
-               case None    => Failure(s"The newly saved change request with ID ${id} was not found back in data base")
+               case None    =>
+                 val msg = s"The newly saved change request with ID ${id} was not found back in data base"
+                 logger.error(msg)
```
fanf: Can you use `ChangeValidationLogger` everywhere ? It would be more consistant for users.

ncharles:
Commit modified

PR #176, opened 2019-07-23
Fixes #15273: Take care of number of nodes in plugin license check

fanf:
https://issues.rudder.io/issues/15273

### DIFF
--- a/api-authorizations/src/main/scala-templates/default/com/normation/plugins/apiauthorizations/EnablePluginImpl.scala
+++ b/api-authorizations/src/main/scala-templates/default/com/normation/plugins/apiauthorizations/EnablePluginImpl.scala
@@ -38,6 +38,7 @@
 package com.normation.plugins.apiauthorizations
 
 import com.normation.plugins.PluginEnableImpl
+import com.normation.rudder.services.nodes.NodeInfoService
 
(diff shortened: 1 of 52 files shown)


fanf:
PR updated with a new commit

api-authorizations/src/main/scala-templates/default/com/normation/plugins/apiauthorizations/EnablePluginImpl.scala
```diff
@@ -38,6 +38,7 @@
 package com.normation.plugins.apiauthorizations
 
 import com.normation.plugins.PluginEnableImpl
+import com.normation.rudder.services.nodes.NodeInfoService
 
-final class CheckRudderPluginEnableImpl() extends PluginEnableImpl
+final class CheckRudderPluginEnableImpl(nodeInfoService: NodeInfoService) extends PluginEnableImpl
```
VinceMacBuche: Do you think in the future we will need to pass RudderConfig directly ? or even directly look into that object ?

fanf: we could. As a general rule, I prefer to let service dependency in parameters, it's easier to test.

PR #205, opened 2019-11-13
Fixes #16074: When deleting completely a rudder-agent, it is very hard to re-install a functionnal agent on it

Fdall:
https://issues.rudder.io/issues/16074


amousset:
Wrong issue

PR #223, opened 2020-01-13
Fixes #16535: Missing scheme error when adding an user from the UI

ElaadF:
https://issues.rudder.io/issues/16535


user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
```diff
@@ -139,20 +136,20 @@ object UserManagementService {
   }
 
   def add(newUser: User): Box[User] = {
+    logger.info(s"path : ${getUserFilePath}")
```
fanf: I don't thing we want that modification

PR #226, opened 2020-01-16
Fixes #16566: Reload users configuration file from the UI

ElaadF:
https://issues.rudder.io/issues/16566


ElaadF:
Commit modified

PR #227, opened 2020-01-17
Fixes #16567: Be able to add a hashed password 

ElaadF:
https://issues.rudder.io/issues/16567

### DIFF
--- a/user-management/src/main/elm/sources/DataTypes.elm
+++ b/user-management/src/main/elm/sources/DataTypes.elm
@@ -8,6 +8,7 @@ module DataTypes exposing (..)
 
 import Dict exposing (Dict)
 import Http exposing (Error)
+import String exposing (toUpper)
 import Toasty
 import Toasty.Defaults
 
@@ -16,6 +17,7 @@ type alias Users = Dict Username Authorization
 type alias Username = String
 type alias Password = String
 type alias RoleConf = List Role
+
 
 type alias Role =
     { id: String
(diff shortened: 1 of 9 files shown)


user-management/src/main/elm/sources/DataTypes.elm
```diff
@@ -65,6 +67,8 @@ type alias Model =
     , userFocusOn : User
     , password : String
     , login : String
+    , hashedPasswd : Bool
+    , clearPasswd : Bool
```
VinceMacBuche: I don't think you need both boolean, I would only keep hashedPassword

user-management/src/main/elm/sources/UserManagement.elm
```diff
@@ -82,11 +79,16 @@ update msg model =
...
 
-        ActivePanelSettings username->
-            ({model | editMod = On, userFocusOn = username}, Cmd.none)
+        ActivePanelAddUser ->
+            if model.addMod == On then
+                ({model | addMod = Off}, Cmd.none)
+            else
+                ({model | addMod = On, editMod = Off}, Cmd.none)
+        ActivePanelSettings username ->
+            ({model | addMod = Off, editMod = On, userFocusOn = username}, Cmd.none)
```
VinceMacBuche: maybe we could have the same thing for to close modification if you click on the same user name

```
if editMod = On and userFocusOn = username
  editMode = Off
else 
  on
```

VinceMacBuche: I think you also need to change you model so you don't have addMod and EditMod but a panelMode of type PanelMode that has 3 value : Closed | AddMode | EditMode String (the username we modify)

VinceMacBuche: This will allow to remove from model the "userFocusOn"

PR #249, opened 2020-03-18
Fixes #16926: Port datasources to zio RC18-2

fanf:
https://issues.rudder.io/issues/16926

### DIFF
--- a/pom-template.xml
+++ b/pom-template.xml
@@ -201,10 +201,10 @@
     <dependency><groupId>commons-fileupload</groupId><artifactId>commons-fileupload</artifactId><scope>provided</scope></dependency>
     <dependency><groupId>commons-io</groupId><artifactId>commons-io</artifactId><scope>provided</scope></dependency>
     <dependency><groupId>commons-lang</groupId><artifactId>commons-lang</artifactId><scope>provided</scope></dependency>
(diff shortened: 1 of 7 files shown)


fanf:
PR updated with a new commit

pom-template.xml
```diff
@@ -201,10 +201,10 @@
     <dependency><groupId>commons-fileupload</groupId><artifactId>commons-fileupload</artifactId><scope>provided</scope></dependency>
     <dependency><groupId>commons-io</groupId><artifactId>commons-io</artifactId><scope>provided</scope></dependency>
     <dependency><groupId>commons-lang</groupId><artifactId>commons-lang</artifactId><scope>provided</scope></dependency>
+    <dependency><groupId>dev.zio</groupId><artifactId>izumi-reflect-thirdparty-boopickle-shaded_${scala-binary-version}</artifactId><scope>provided</scope></dependency>
```
VinceMacBuche: Why do we need to adjust dependencies on plugins-common ?

fanf: they changed some dependencies when switching to zio `RC18-2`. 
Maven don't manage the provided one correctly if we don't tell him like that, so it may put them in the rpkg (because it doesn't see that they are provided), which lead to bigger rpkg (and horror stories in case of version mismatch)

PR #252, opened 2020-03-23
Fixes #16952: error at installation of openscap plugin

ncharles:
https://issues.rudder.io/issues/16952


ncharles:
OK, merging this PR

PR #256, opened 2020-03-24
Fixes #16962: correct path to openscap node-external-report config file

ncharles:
https://issues.rudder.io/issues/16962


fanf:
Given that we decided to not rename `openscap-report` for branche 5.0, we should close that one and corresponding ticket, right?

fanf:
Closing that one because we swithed to 6.1 for CIS.

PR #263, opened 2020-03-30
Fixes #16997: Rename plugin directory

Fdall:
https://issues.rudder.io/issues/16997


ncharles:
it's been done in 6.1

Fdall:
Closing since it is not needed anymored. First version of the plugin should be in 6.1

PR #271, opened 2020-04-07
Fixes #16983: display openscap report in node details

ncharles:
https://issues.rudder.io/issues/16983

### DIFF
--- a/openscap-policies/src/main/scala/com/normation/plugins/openscappolicies/api/OpenScapApi.scala
+++ b/openscap-policies/src/main/scala/com/normation/plugins/openscappolicies/api/OpenScapApi.scala
@@ -1,0 +1,133 @@
+package com.normation.plugins.openscappolicies.api
+
+import java.io.InputStream
+
+import com.normation.inventory.domain.NodeId
+import com.normation.plugins.openscappolicies.OpenscapPoliciesLogger
+import com.normation.plugins.openscappolicies.services.{OpenScapReportReader, ReportSanitizer}
+import com.normation.rudder.api.HttpAction.{GET, PUT}
+import com.normation.rudder.rest.EndpointSchema.syntax._
+import com.normation.rudder.rest.{ApiModuleProvider, EndpointSchema, GeneralApi, SortIndex, StartsAtVersion10, ZeroParam}
+import com.normation.rudder.rest.EndpointSchema.syntax._
+import com.normation.rudder.rest._
+import com.normation.rudder.rest.lift.{DefaultParams, LiftApiModule, LiftApiModule0, LiftApiModuleProvider}
+import net.liftweb.http.{InMemoryResponse, LiftResponse, Req}
+import net.liftweb.json.{JValue, NoTypeHints}
+import net.liftweb.common.{Box, EmptyBox, Failure, Full}
+import sourcecode.Line
+import com.normation.box._
+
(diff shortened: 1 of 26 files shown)


ncharles:
style is not finished (and far from it)
missing the policy part to generate the report
idealy: async loading of the frame

ncharles:
Commit modified

openscap-policies/src/main/scala/com/normation/plugins/openscappolicies/extension/OpenScapNodeDetailsExtension.scala
```diff
@@ -0,0 +1,87 @@
...
+      "popupDetails" -> addOpenScapReportTab(snippet) _
+    , "mainDetails"  -> addOpenScapReportTab(snippet) _
+  )
+
+  /**
+   * Add a tab:
+   * - add an li in ul with id=openScapExtensionTab
+   */
+  def addOpenScapReportTab(snippet: ShowNodeDetailsFromNode)(xml: NodeSeq): NodeSeq = {
+    val content = openScapReader.checkOpenScapReportExistence(snippet.nodeId) match {
```
fanf, line 31: here, perhaps we should check if the node has an openscap directive, and only in that case show the tab (and check report existence) ?

openscap-policies/src/main/scala/com/normation/plugins/openscappolicies/extension/OpenScapNodeDetailsExtension.scala
```diff
@@ -0,0 +1,87 @@
...
+              <div class="error">No OpenScap report available for node {snippet.nodeId.value}</div>
+            </div>
+          case true =>
+            frameContent(snippet.nodeId)(openScapExtensionXml)
+        }
+    }
+
+
+      //tabContent(Map("key" ->"value"))(openScapExtensionXml)
+    val tabTitle = "OpenScap"
```
fanf, line 49: open scap write it `OpenSCAP`, we should keep that typography for user facing things

openscap-policies/src/main/scala/com/normation/plugins/openscappolicies/api/OpenScapApi.scala
```diff
@@ -0,0 +1,133 @@
...
+          logger.trace("Doing in memory response")
+          InMemoryResponse(
+            sanitizedReport.toString().getBytes()
+            // somehow, the X-FRAME-OPTIONS get rewrote to DENY here
+            ,  ("Content-Type", "text/html") :: ("Content-Disposition", "inline") :: Nil
+            , Nil
+            , 200)
+        case eb: EmptyBox =>
+          val errorMessage = eb ?~! "Could not get the sanitized OpenScap report for node ${nodeId}"
+          logger.error(errorMessage.messageChain)
```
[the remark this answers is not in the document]
fanf: hum, in fact it won't work in your case, because that API call is use only internally to display the result directly. 
So that API should really be only internal I think

ncharles:
Commit modified

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
Commit modified

ncharles:
PR updated with a new commit

ncharles:
this is targeting 6.0 rather than 6.1. Closing

PR #274, opened 2020-04-08
Vault cert

ikramBej:


vault/share/plugins/vault/20_cfe_basics/vault_lib.cf
```diff
@@ -29,6 +29,7 @@
 #       "password": "PASSWD"
 #     }
 #   }
```
amousset, line 31: missing comma here

vault/share/plugins/vault/sample_vault.json
```diff
@@ -9,4 +9,5 @@
       "password": "PASSWD"
     }
   }
```
amousset, line 11: missing comma here

vault/share/plugins/vault/sample_vault.json
```diff
@@ -9,4 +9,5 @@
...
+    "verify_certificate": ""
```
amousset, line 12: default value is `true`

vault/share/plugins/vault/20_cfe_basics/vault_lib.cf
```diff
@@ -38,6 +39,16 @@ bundle agent vault_fetch_config()
   vars:
       "config_path" string => "/var/rudder/plugin-resources/vault.json";
       "vault.config" data => readjson("${config_path}");
+ 
+      "vault.cert" string => "--insecure",
+      ifvarclass => strcmp("vault.config[verify_certificate]", "true"),
```
amousset, line 44: no need for is variable here is is strcmp works it means the variale exists

amousset: and `--insecure` means we do *not* verify the certificates, the logic is inverted here

PR #277, opened 2020-04-08
Fixes #17086: Missing scheme error when adding an user 

ElaadF:
https://issues.rudder.io/issues/17086


ElaadF:
OK, merging this PR

PR #279, opened 2020-04-09
fixes #17042: Plugins should not use a hardcoded path for curl

ikramBej:
https://issues.rudder.io/issues/17042


amousset:
OK, merging this PR

PR #300, opened 2020-05-12
Fixes #17387: Add more item support for debian9 benchmark

ico12345:
https://issues.rudder.io/issues/17387


VinceMacBuche:
Sorry for the very late reply, cis plugin si now deprecated. It may come back in a distant future, with a wider supported os scope but for now it will not happen

PR #307, opened 2020-05-29
Fixes #17570: Trigger promises generation when promoting a node to relay

ElaadF:
https://issues.rudder.io/issues/17570

### DIFF
--- a/scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
+++ b/scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
@@ -6,6 +6,8 @@ import com.normation.errors._
 import com.normation.eventlog.EventActor
 import com.normation.eventlog.ModificationId
 import com.normation.inventory.domain.NodeId
+import com.normation.rudder.batch.AsyncDeploymentActor
+import com.normation.rudder.batch.AutomaticStartDeployment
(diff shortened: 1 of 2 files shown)


ElaadF:
Commit modified

VinceMacBuche:
Are you sure it needs a reload ? i don't think because rule creation already reloads

ElaadF:
I don't see where the reload is performed, no eventlog indicates this and it seems that in https://github.com/Normation/rudder/blob/master/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala#L186 there is no reload at this point

scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -59,6 +61,7 @@ class ScaleOutRelayService(
                     } else {
                       nodeInfos.succeed
                     }
+      _ <- clearCacheService.clearNodeConfigurationCache(storeEvent = true, actor)
```
VinceMacBuche, line 64: In fact you should trigger a deployment not delete the whole configurationCache.

We used to call deploy api that only starts a deployment:

```
      asyncDeploymentAgent ! AutomaticStartDeployment(modId, actor)
```

with 
```
  , asyncDeploymentAgent : AsyncDeploymentActor
```

you can put it in the yield section

ElaadF:
Commit modified

PR #322, opened 2020-06-29
Bug 17866/add screenshots for validated users

ElaadF:
https://issues.rudder.io/issues/17866


scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -212,6 +212,9 @@ class ScaleOutRelayService(
      for {
       authorizedNetworks <- policyServerManagementService.getAuthorizedNetworks(nodeInf.policyServerId)
     } yield {
+       println(authorizedNetworks)
```
amousset, line 215: I guess this should not be part of the PR.

PR #341, opened 2020-10-22
Fixes #18394:  File method isWriteable has become isWritable

ElaadF:
https://issues.rudder.io/issues/18394


ElaadF:
Should go in 6.2

PR #350, opened 2020-11-19
Fixes #18643: Datasources test in error

fanf:
https://issues.rudder.io/issues/18643

### DIFF
--- a/datasources/src/main/scala/com/normation/plugins/datasources/Scheduler.scala
+++ b/datasources/src/main/scala/com/normation/plugins/datasources/Scheduler.scala
@@ -87,13 +87,14 @@ class DataSourceScheduler(
       case Scheduled(d)  =>
         if(datasource.enabled) {
           DataSourceLoggerPure.Scheduler.info(s"Datasource '${datasource.name.value}' (${datasource.id.value}) is enabled and scheduled every ${d.asScala.toMinutes.toString} minutes") *>
-          Schedule.spaced(d).succeed
+          // This historical semantic is "do a sync immediately and then one spaced every 'd'"
+          Schedule.once.andThen(Schedule.spaced(d)).succeed
         } else {
           DataSourceLoggerPure.Scheduler.info(s"Datasource '${datasource.name.value}' (${datasource.id.value}) is disabled") *>
           never.succeed
         }
       case NoSchedule(_) => //in that case, our source doesn't produce anything
-        DataSourceLoggerPure.Scheduler.info(s"Datasource '${datasource.name.value}' (${datasource.id.value}) is enabled and but no schedule is configured") *>
+        DataSourceLoggerPure.Scheduler.info(s"Datasource '${datasource.name.value}' (${datasource.id.value}) is enabled but no schedule is configured") *>
         never.succeed
     }
 
@@ -108,7 +109,7 @@ class DataSourceScheduler(
 
     for {
       s <- schedule
-      p <- prog.repeat(s).provide(clock).unit
+      p <- prog.schedule(s).provide(clock).unit
     } yield {
       p
     }
--- a/datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
+++ b/datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
@@ -482,11 +482,11 @@ class UpdateHttpDatasetTest extends Specification with BoxSpecMatcher with Logga
   }
 
   implicit class QueueFailIfNonEmpty[A](queue: Queue[A]) {
-    def failIfNonEmpty = queue.poll.map(_.map(_ => ().fail))
+    def failIfNonEmpty: IOResult[Unit] = queue.poll.flatMap {
+      case None    => ().succeed
+      case Some(_) => Inconsistency(s"queue should be empty but size = ${1+queue.size.runNow}").fail
+    }
   }
-
-
-  sequential
 
   "Array validation with [*]" >> {
     Fragment.foreach(0 until testArray.size) { i =>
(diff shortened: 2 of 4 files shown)


datasources/src/main/scala/com/normation/plugins/datasources/Scheduler.scala
```diff
@@ -108,7 +110,7 @@ class DataSourceScheduler(
 
     for {
       s <- schedule
-      p <- prog.repeat(s).provide(clock).unit
+      p <- prog.schedule(s).provide(clock).unit
```
fanf, line 113: We need to use `schedule` to make `never` work, because `repeat` always execute at least once while `schedule` starts by trying to understand the schedule. That's also why we need to use `Schedule.once.andThen` now, we don't get the one free from `repeat`. Not that the semantic is better now, `repeat` always repeats, and there is no magical things done for `Schedule.never` to work with it.

fanf:
PR updated with a new commit

datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
```diff
@@ -663,31 +665,31 @@ class UpdateHttpDatasetTest extends Specification with BoxSpecMatcher with Logga
           _        <- dss.restartScheduleTask()
           //then just after, we have the first exec - it still need at least a ms to tick
           //still nothing here
-          _        <- testClock.get[TestClock.Service].setTime(1 millis)
+          _        <- testClock.get[TestClock.Service].adjust(1.second)
```
fanf, line 668: `setTime` doesn't make the test clock advance :/

datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
```diff
@@ -822,56 +824,14 @@ class UpdateHttpDatasetTest extends Specification with BoxSpecMatcher with Logga
...
-              }
-          ]}""").getValue("x"))):Option[NodeProperty]))
+          Some(DataSource.nodeProperty("test-get-one-node",  ConfigFactory.parseString("""{ "x" :
+          {
+              "author" : "Nigel Rees",
+              "category" : "reference",
+              "price" : 8.95,
+              "title" : "Sayings of the Century"
+          }
+          }""").getValue("x"))):Option[NodeProperty]))
```
fanf, line 834: We use [:1] in the test, there is only one object returned, that's the goal

datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
```diff
@@ -1026,16 +986,16 @@ class UpdateHttpDatasetTest extends Specification with BoxSpecMatcher with Logga
...
   *   - so an array of array will return an array
   * - an array with several values => keep array
-  */ 
+  */
   // testing arrays, array of arrays, etc
   lazy val testArray = List(
     // [] is a special case handled elsewhere
     // what API give                             -   expected with [*]                          - expected with [:1]
     ("""[ 1 ]"""                                 , """1"""                                      , """1"""                                 )
-  , ("""[ "a" ]"""                               , """"a""""                                    , """"a""""                               )
```
fanf, line 1035: the string becomes top-level, so we unquote it

datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
```diff
@@ -1026,16 +986,16 @@ class UpdateHttpDatasetTest extends Specification with BoxSpecMatcher with Logga
...
+  , ("""[ "a" ]"""                               , """a"""                                      , """a"""                                 )
   , ("""[ { "a" : 1 }]"""                        , """{"a":1}"""                                , """{"a":1}"""                           )
-  , ("""[ 1, 2 ]"""                              , """[1,2]"""                                  , """1"""                                 ) // array of size 1 are lifted
-  , ("""[ "a", "b" ]"""                          , """["a","b"]"""                              , """"a""""                               ) // array of size 1 are lifted
+  , ("""[ 1, 2 ]"""                              , """["1","2"]"""                              , """1"""                                 ) // array of size 1 are lifted
```
fanf, line 997: this one... I'm not sure we want to keep it like that. Our top object are only either json object, json arrays, or json string. Here, we look at each int, transform it in a valid top object (so a string), and put it back in an array. I'm pretty sure it will breaks things, but I'm not sure how to deal with that.

VinceMacBuche: I think we should keep the type we have (int, boolean)

Maybe we should look how the agent treats int / boolean when read from a json property

VinceMacBuche: So we could have the more suitable format for this data

ncharles: I'm pretty sure we should not transform int to string. Why can't we have a top object json int ?

VinceMacBuche: anyway for now I'll just merge it and we need to think what we want to achieve here (and it will be on Rudder / properties side)

VinceMacBuche (approved):

datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
fanf: @ncharles that would break compat. Historically, we only had string, but without quote: if an user put `1`, internally for us it's the string `1`, but it's outputed as `1` whitout quotes. It means that we don't put semantic besides string for that, and it's because we don't understand anything but string for directive parameters. 
We added JSON object and array at top level, and in the only case where we do extraction for the user (access of properties in `${node.properties[]}`), we take care to remove quotes on string. 

I have an idea to not break it.

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #352, opened 2020-12-09
Fixes #18757: Add more item support for debian9 and debian10 benchmark

ico12345:
https://issues.rudder.io/issues/18757


itacos:
Hi,

I just pushed raw export of debian9 and debian10 rules because I have some problems when using the parser.


It's difficult to work with multiple benchmark at the same time because we need to respect the pattern "CIS - Item title" so we can not specify the associated benchmark.


Also technique names are associated to a benchmark but in fact they can be used in multiple benchmark.


I had duplicated lot of techniques and directives so I think it would be better to regen/check uuids.


Thanks

VinceMacBuche:
Sorry for the very late reply, cis plugin si now deprecated. It may come back in a distant future, with a wider supported os scope but for now it will not happen

PR #360, opened 2021-04-14
Fixes #19147: Email notification should be disabled when no host server is configured

ElaadF:
https://issues.rudder.io/issues/19147

### DIFF
--- a/scale-out-relay/src/main/resources/server-node-to-relay
+++ b/scale-out-relay/src/main/resources/server-node-to-relay
@@ -29,7 +29,7 @@ else
   CERTIFICATE_OPTION="--insecure"
 fi
 
-DOWNLOAD_COMMAND="curl --silent --show-error ${CERTIFICATE_OPTION} --location --noproxy --globoff"
+DOWNLOAD_COMMAND="curl --silent --show-error ${CERTIFICATE_OPTION} --location --proxy '' --globoff"
 
 usage() {
 	echo "Usage: $0 <UUID of the node to make a relay server>"


change-validation/src/main/scala/com/normation/plugins/changevalidation/NotificationService.scala
```diff
@@ -128,13 +129,17 @@ class NotificationService(
...
     for {
       serverConfig  <- getSMTPConf(configMailPath)
-      emailConf     <- getStepMailConf(step, configMailPath)
-      rudderBaseUrl <- getRudderBaseUrl(configMailPath)
-      params        =  extractChangeRequestInfo(rudderBaseUrl, cr)
-      mf            =  new DefaultMustacheFactory()
-      emailBody     <- getContentFromTemplate(mf, emailConf, params)
-      emailSubject  <- getSubjectFromTemplate(mf, emailConf.subject, params)
-      _             <- emailService.sendEmail(serverConfig, emailConf.toEnvelop(emailBody).copy(subject = emailSubject))
+      _             <- ZIO.when(!serverConfig.smtpHostServer.isBlank) {
```
VinceMacBuche, line 132: isEmpty not isBlank (we still support some old Java8 systems that does not support isBlank

ElaadF:
Commit modified

change-validation/src/main/scala/com/normation/plugins/changevalidation/NotificationService.scala
```diff
@@ -128,13 +129,17 @@ class NotificationService(
...
       serverConfig  <- getSMTPConf(configMailPath)
-      emailConf     <- getStepMailConf(step, configMailPath)
-      rudderBaseUrl <- getRudderBaseUrl(configMailPath)
-      params        =  extractChangeRequestInfo(rudderBaseUrl, cr)
-      mf            =  new DefaultMustacheFactory()
-      emailBody     <- getContentFromTemplate(mf, emailConf, params)
-      emailSubject  <- getSubjectFromTemplate(mf, emailConf.subject, params)
-      _             <- emailService.sendEmail(serverConfig, emailConf.toEnvelop(emailBody).copy(subject = emailSubject))
+      _             <- ZIO.when(serverConfig.smtpHostServer.nonEmpty) {
+                         for {
```
fanf, line 133: I think there should be an info log line in each case - oh, no the config is read on each notification. 
OK, we need to add the info in the plugin page in a next version. (ie in the plugin page, something like "Email notification are [disabled (configure email server in /... to enable) | enable (comment config in .... to disable) ]

fanf (changes requested):
You need to update documentation of the plugin to reflect that change in behavior

change-validation/src/main/scala/com/normation/plugins/changevalidation/NotificationService.scala
ElaadF: Do you mean we want to log whether active or not when we try to send an email ?

fanf: yes, but I forgot that we reread conf each time, so it would hard to spot when it changes from enable to disable (or opposite). 
So in place of having a log when the config is read, we need to display the info somewhere else.

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Moved to #364

PR #367, opened 2021-04-30
Fix #19206: Missing cron task for application of monitoring templates

juhnny5:


zabbix/packaging/postinst
```diff
@@ -1,13 +1,20 @@
...
-
+  
 CONFFILE="/opt/rudder/etc/zabbix.conf"
+CRONFILE="/etc/cron.d/zabbix-rudder"
 
 mkdir -p /var/rudder/plugin-resources
-echo "0 * * * * /opt/rudder/bin/zabbix-plugin.py update " > /etc/cron.d/zabbix-rudder
+
+# create cron file
+if [ ! -f "${CRONFILE}"]; then
```
amousset, line 9: we should probably replace it, I don't know why this test. Do you know @peckpeck?

peckpeck: it was to avoid changing user modified cron, in some case it is probably better to replace it

peckpeck:
PR replaced by https://github.com/Normation/rudder-plugins/pull/716

PR #376, opened 2021-05-27
Fixes #19252: 

ncharles:
https://issues.rudder.io/issues/19252

### DIFF
--- a/openscap/src/main/resources/antisamy.xml
+++ b/openscap/src/main/resources/antisamy.xml
@@ -51,12 +51,12 @@ http://www.w3.org/TR/html401/struct/global.html
 				
 		<regexp name="anything" value=".*"/>
 		<regexp name="numberOrPercent" value="(\d)+(%{0,1})"/>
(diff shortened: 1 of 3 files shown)


openscap/src/main/resources/antisamy.xml
```diff
@@ -85,7 +85,7 @@ http://www.w3.org/TR/html401/struct/global.html
 		<regexp name="cssPseudoElementSelector" value=":[a-zA-Z0-9\-_\(\)\:]+"/>
 
 		<!--  This is to list out any psuedo-element names that are *not* valid -->
-		<regexp name="cssPseudoElementExclusion" value=""/>
+		<regexp name="cssPsuedoElementExclusion" value=""/>
```
VinceMacBuche, line 88: here is a typo on the name did you wrote the file @ncharles or it came with the dependency ?

ncharles: it cale with the dependency :(

fanf (approved):

PR #380, opened 2021-06-07
Fixes #19229: Deadlock at start when loading new licence

fanf:
https://issues.rudder.io/issues/19229

So, I wasn't able to reproduce the problem, and that PR is a bit of testing from what I know can lead to that given the thread dump @ncharles got. 

My best guess is that the `lazy val` in the object + `runNow` is the root of the problem. We had that problem (hard to debug) in other tickets (https://issues.rudder.io/issues/17228). 

So: change the `lazy val` for `cache`. 

There was other problems in the services init that are unlikely to be linked to the deadlock, but who knows?

- don't use hardcoded path (moreover, it can be a problem in security audit), use them in class parameters;
- there was a class for whose package declaration and directory tree was not matching, which is known to lead to strange things in the JVM/Classloaders. 
- there was duplicated resources (in `toserve/branding` and `branding/toserve`),


branding/src/main/scala/com/normation/plugins/branding/BrandingConfService.scala
```diff
@@ -73,11 +73,13 @@ class BrandingConfService extends Loggable {
...
 
+  val defaultConfigFilePath = "/var/rudder/plugins/branding/configuration.json"
 
-  val configFilePath = "/var/rudder/plugins/branding/configuration.json"
+}
 
+class BrandingConfService(configFilePath: String) {
 
-  private[this] lazy val cache : Ref[Either[RudderError, BrandingConf]] = (for {
+  private[this] val cache : Ref[Either[RudderError, BrandingConf]] = (for {
```
fanf, line 82: this is likely the culprit

ncharles (approved):

fanf:
OK, merging this PR

PR #403, opened 2021-08-23
Fixes #19844: New changes to plugin build

VinceMacBuche:
https://issues.rudder.io/issues/19844


makefiles/make-plugins-common.mk
```diff
@@ -32,4 +38,4 @@ $(PARENT_POM):
 	$(MVN_CMD) install
 clean:
 	rm -f  pom.xml
-	rm -rf target
+		rm -rf target
```
amousset, line 41: indentation typo

api-authorizations/pom-template.xml
```diff
@@ -11,7 +11,7 @@
   <parent>
     <groupId>com.normation.plugins</groupId>
     <artifactId>plugins-parent</artifactId>
-    <version>${rudder-branch}-${parent-version}</version>
+    <version>${parent-version}</version>
```
fanf, line 14: I don't understand that. Plugin parent must be by branch, since it holds all dependencies. Do you want to change the way it is versionned ?

fanf: In fact, I don't understand at all. The sed that used to replace it by the plugin-parent version is deleted, and I don't understand where it comes from now.

VinceMacBuche:
Commit modified

fanf (commented):
Can you please explain somewhere (either in the description here, or in a a private doc on notion, or in the readme, or in the issue) what is the chosen logic and convention and how it will work in 7.0 ?

api-authorizations/pom-template.xml
VinceMacBuche: it's still by branch, it just that i refactored to have less variables and they are now treated by the makefile depending on the target called

VinceMacBuche: see here

VinceMacBuche: See here: https://github.com/Normation/rudder-plugins/pull/403/files#diff-a288abe3cd89222034a3c163912d7d6769a3bddd75d2223b51ca02d92386cceeR28

VinceMacBuche: I just removed useless variables (like nightly variables) and now the nightly and snapshot are put on the right place whether you call the correct make target ( all / nightly or next) the makefille will now create a put the right value at the right place instead of depending on a value from a conf file that was either wrong or needed to be manipulated

VinceMacBuche:
The changes are that we now

* generate a rpkg on nightly that does not use the POM_VERSION (which used to contain SNAPSHOT within 
https://github.com/Normation/rudder-plugins/pull/403/files#diff-a288abe3cd89222034a3c163912d7d6769a3bddd75d2223b51ca02d92386cceeR28

* and also that the version used in metadata with nightlies does not containt SNAPSHOT either (same way by not using POM_VERSION but just VERSION, no more metadata-nightly target) https://github.com/Normation/rudder-plugins/pull/403/files#diff-a288abe3cd89222034a3c163912d7d6769a3bddd75d2223b51ca02d92386cceeR28

api-authorizations/pom-template.xml
fanf: ok, the definition part was missing in the revesion I reviewed, now I understand, thanks.

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #405, opened 2021-08-25
Fixes #19859: Fix typos in datasources plugins

amousset:
https://issues.rudder.io/issues/19859

### DIFF
--- a/datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
+++ b/datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
@@ -55,10 +55,10 @@ import zio.syntax._
 import zio.duration.DurationOps
 
 /*
- * This file contain the hight level logic to update
+ * This file contain the high level logic to update
  * datasources by name:
  * - get the datasource by name,
- * - get the list of nodes to udpate and there context,
+ * - get the list of nodes to update and there context,
  * - update all nodes
  * - (event log are generated because we are just changing node properties,
  *   so same behaviour)
@@ -244,7 +244,7 @@ class HttpQueryDataSourceService(
     for {
       mode          <- globalPolicyMode()
       updated       <- tasks(nodes, info.policyServers, mode, info.parameters).timeout(timeout).provide(clock).notOptional(
-                         s"Timout error after ${timeout.asScala.toString()}"
+                         s"Timeout error after ${zio.duration.Duration.fromJava(timeout).render}"
                        )
                        // execute hooks
       _             <- onUpdatedHook(updated.collect { case Right(NodeUpdateResult.Updated(id)) => id }.toSet, cause)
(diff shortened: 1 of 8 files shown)


datasources/src/main/scala/com/normation/plugins/datasources/Serialization.scala
```diff
@@ -417,7 +417,7 @@ trait DataSourceExtractor[M[_]] extends JsonExctractorUtils[M] {
 }
 
 object DataSourceExtractor {
-  object OptionnalJson extends DataSourceExtractor[Option] {
+  object OptionalJson extends DataSourceExtractor[Option] {
```
amousset, line 420: Is it really ok to rename it? I'm not sure it canot be used somewhere else

datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
```diff
@@ -244,7 +244,7 @@ class HttpQueryDataSourceService(
     for {
       mode          <- globalPolicyMode()
       updated       <- tasks(nodes, info.policyServers, mode, info.parameters).timeout(timeout).provide(clock).notOptional(
-                         s"Timout error after ${timeout.asScala.toString()}"
+                         s"Timeout error after ${timeout.asScala.toString()}"
```
fanf, line 247: use `.render` in place of `toString` to have an human readable format

amousset:
PR updated with a new commit

datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
amousset: thanks!

datasources/src/main/scala/com/normation/plugins/datasources/Serialization.scala
fanf: you need to change it where it's used, too (like you did) but appart that, it's ok

fanf: since datasources is a plugin (and plugin can't depend of other plugin for now), it's safe

amousset: :+1:

datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
```diff
@@ -244,7 +244,7 @@ class HttpQueryDataSourceService(
     for {
       mode          <- globalPolicyMode()
       updated       <- tasks(nodes, info.policyServers, mode, info.parameters).timeout(timeout).provide(clock).notOptional(
-                         s"Timout error after ${timeout.asScala.toString()}"
+                         s"Timeout error after ${timeout.asScala.render()}"
```
fanf, line 247: so it seems that there is 12 different durations, not all with render, let me see what is happening

fanf: Ok, it's that one: ```s"Timout error after ${zio.duration.Duration.fromJava(timeout).render}"```

amousset:
PR updated with a new commit

fanf (approved):

amousset:
OK, squash merging this PR

PR #425, opened 2021-10-18
Fixes #20135: Missing -X POST in request to reload techniques in node-to-relay

amousset:
https://issues.rudder.io/issues/20135


scale-out-relay/src/main/resources/server-node-to-relay
```diff
@@ -69,12 +69,12 @@ if [ "${result_api}" = "error" ]; then
 	echo "Reason: ${error_details}"
 	exit 1
 elif [ ${parse_response} = 4 ];then
-	echo "ERROR: Unknow response format, parsing failed. Code ${parse_response}" 1>&2
```
VinceMacBuche, line 72: poor typo, destroyed here :(

PR #428, opened 2021-11-15
Fixes #19703: openscap display is broken in 7.0

ncharles:
https://issues.rudder.io/issues/19703


ncharles:
it looks like that
image
there is an extra spacing, but I can't find why - generated HTML is the same than for "node state" tab

VinceMacBuche (approved):

PR #430, opened 2021-12-01
Fixes #20353: Change default color to match the new theme

amousset:


branding/src/main/elm/sources/Branding.elm
```diff
@@ -26,7 +25,8 @@ port applyCss : ( CssObj ) -> Cmd msg
 defaultSettings : Settings
 defaultSettings =
   let
-    bgColor  = Color.red
+    --- #da291c, from 7.0 color palette
+    bgColor  = rgb255 218 41 28
```
ncharles, line 29: i'm fairly sure the defaut colour needs to be changed in com/normation/plugins/branding/BrandingConfService.scala as well - inital color is JsonColor(204,0,0,1)

VinceMacBuche (approved):

PR #447, opened 2022-02-16
Fixes #20777: Adapt plugin to change in UserDetailsService

fanf:
https://issues.rudder.io/issues/20777


fanf:
Duplicate of #445

PR #451, opened 2022-03-14
Fixes #20778: Add Oauth2/openid connect backend

fanf:
https://issues.rudder.io/issues/20778

Add the possibility to have an Oauthv2/OIDC (openid connect) SSO authentication in Rudder. 
Under the hood, it uses spring-security module for Oauthv2 (https://docs.spring.io/spring-security/reference/5.7/reactive/oauth2/index.html). 
The main difficulties is that contrary to other existing backend, Oauthv2 is a change in the list of security filter chain, not a new authentication "data base" for the web authentication filter. 
So we need to hack existing chain (configured in main rudder) and hand-build missing filters for oauth/oidc and but them back at the correct place. This is complicated. It's done in `AuthBackendsConf.scala` in `AuthBackendsSpringConfiguration`.

In that class: 
- we read and update existing filter chain in `setApplicationContext` method (spring method for "do stuff with spring stuff, even the goriest one that would not be allowed elsewhere")
- all `@Bean` below are just the need plumbing to build by hand what spring would do with the `<oauth>` smart tag, that we can't use, because we can't "inject" xml config file once the main `<http>` chain is built (we can just add new `<http>` chain, which is useless)
Even with that, we still need `applicationContext-security-auth-oidc.xml` and `applicationContext-security-auth-oauth2.xml` because it's these files that allows our licence system to know if the corresponding "auth backend" are enabled (so, ~ useless for the logic of authenticating users with oauth, needed for plumbing with license check).

We also need to extends what is a `userDetails` because oauth has its own with specific infor related to the protocol (some info can be shared between the identity provider and an application like rudder). 

`Oauth2Authentication.scala` is just for parsing the complicated configuration needed for oauth and build an easy to deal with data structure. You can see an attempt to explain needed configuration properties in the `README.adoc`.

Finally, when we have an Single Sign On solution, we likely don't want to show to standard rudder login form, since, well, you're supposed to log on the identity provider site. But perhaps you still want to show it for emergency (like "no more network with the identity provider"). So there's a new config option with the plugin to either `show`, `hide` (ie hide but available with a toggle button), or `remove` (ie: don't sent the HTML at all) Rudder form.

### DIFF
--- a/auth-backends/README.adoc
+++ b/auth-backends/README.adoc
@@ -12,7 +12,23 @@ other general information.
 
 = Authentication backends
 
-This plugins allows to use alternative authentication backends for Rudder: LDAP/AD and radius.
+This plugins allows to use alternative authentication backends for Rudder: SAMLv2/OpenID Connect, LDAP/AD and radius (DEPRECATED).
+
+Each authentication method is detailed below. Users are expected to know how an authentication system works independently from Rudder to configure it in Rudder: you will likely need authentication token, URLs, and other properties provided by your company.
+
+== Configure login form rendering
+
+By default, the standard Rudder login form is displayed. SSO backends like OpenID Connect ones add links to the relevant SSO own login page below that Rudder login form. When you use such an authentication method, you want to hide or totaly remove Rudder own login form to avoid to confuse you user. For that, you can use the following property:
+
+```
+rudder.auth.displayLoginForm=show
+```
+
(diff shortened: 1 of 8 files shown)


fanf:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche (approved):

PR #461, opened 2022-04-14
Fixes #20997: Incorrect error of missing oauth config when the backend is disabled 

fanf:
https://issues.rudder.io/issues/20997

The 3 following cases are now working correctly:
- when oicd/oauth is not configured, then login snippet is not registered and spring security are not created, 
- when oicd/oauth is configured in property file by properties are missing, we have the error message on login page, 
- when everything is correctly configured, no error message (and oidc auth works). 

In all cases, fallback authentication was working. 

From a tech point of view, it's a bit ugly, because we don't have our hand on what is the exact init graph that spring uses, and we don't want to have to change things in rudder core, and we have plugins which complicated things. So we just read properties file also in the plugin just to know if the user want oidc/oauth, and if not, we don't init related things.


VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #486, opened 2022-07-18
Fixes #21428: output of install of scale out relay is clearly not great

ncharles:
https://issues.rudder.io/issues/21428


amousset:
THANK YOU

fanf (approved):

PR #501, opened 2022-09-26
Fixes #21826: Port plugins to ZIO2

fanf:
https://issues.rudder.io/issues/21826

Same general changes than https://github.com/Normation/rudder/pull/4507 but in addition, some more related to Datasources. We use since rudder 7.2 `zhttp` for the tests which simplifies a lot the cats dependencies spagghetti ball. The bad point is that for `ZIO 2`, we need to use `2.0.0.x` version, which happens to have a breaking change in the semantic of URL pattern matching. It caused some troubles to be identified. Most of the other changes in [UpdateHttpDatasetTest.scala](https://github.com/Normation/rudder-plugins/pull/501/files#diff-44d43b7b8b22fabae4537dae04dcf91ea30fe5594278681a49ffe9e43951e4e8) 
Are related to either reodering of code, some cleaning (partly thanks to better ZIO features and partly because the dev are better with it). 
The removal of test clock worked, which is suprising.

### DIFF
--- a/auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
+++ b/auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
@@ -284,7 +284,7 @@ class AuthBackendsSpringConfiguration extends ApplicationContextAware {
       } yield {
         r.toMap
       }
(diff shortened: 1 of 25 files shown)


fanf:
PR updated with a new commit

PR #516, opened 2022-10-21
Fixes #22004: Spotless must not be exec in plugin parent

fanf:
https://issues.rudder.io/issues/22004


fanf:
OK, merging this PR

PR #517, opened 2022-10-22
Fixes #22008: Spotless must not be exec in plugins-common but on children

fanf:
https://issues.rudder.io/issues/22008


fanf:
OK, merging this PR

PR #519, opened 2022-11-06
Fixes #22052:  Update elm dependencies - plugins

amousset:
https://issues.rudder.io/issues/22052

### DIFF
--- a/api-authorizations/src/main/elm/elm.json
+++ b/api-authorizations/src/main/elm/elm.json
@@ -6,14 +6,14 @@
     "elm-version": "0.19.1",
     "dependencies": {
         "direct": {
-            "NoRedInk/elm-json-decode-pipeline": "1.0.0",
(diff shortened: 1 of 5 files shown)


auth-backends/src/main/elm/sources/AuthBackends.elm
```diff
@@ -428,15 +417,8 @@ getErrorMessage e =
     let
         errMessage =
             case e of
-                Http.BadStatus b ->
```
amousset, line 431: This is a regression, but getting the request content extends my current elm skill.

amousset:
PR updated with a new commit

auth-backends/src/main/elm/sources/AuthBackends.elm
amousset: Ok, `Http.Detailed`

VinceMacBuche (approved):

PR #525, opened 2022-11-30
Fixes #22158: LDAP auth backend should deref link to allows finding user in remote directory

fanf:
https://issues.rudder.io/issues/22158


fanf:
Waiting for user confirmation that it solves the authentication problem

fanf:
The root cause was something else, but I think that change is still valuable. It sould be exposed in a property though

fanf:
PR updated with a new commit

fanf:
OK, merging this PR

PR #531, opened 2023-01-20
Fixes #22188: Changes made by validated users still need to be validated

ElaadF:
https://issues.rudder.io/issues/22188

### DIFF
--- a/change-validation/src/main/scala/bootstrap/rudder/plugin/ChangeValidationConf.scala
+++ b/change-validation/src/main/scala/bootstrap/rudder/plugin/ChangeValidationConf.scala
@@ -59,7 +59,6 @@ import com.normation.plugins.changevalidation.RoWorkflowJdbcRepository
 import com.normation.plugins.changevalidation.SupervisedTargetsReposiory
 import com.normation.plugins.changevalidation.TopBarExtension
 import com.normation.plugins.changevalidation.TwoValidationStepsWorkflowServiceImpl
-import com.normation.plugins.changevalidation.UserValidationNeeded
 import com.normation.plugins.changevalidation.ValidatedUserMapper
 import com.normation.plugins.changevalidation.ValidationNeeded
 import com.normation.plugins.changevalidation.WoChangeRequestJdbcRepository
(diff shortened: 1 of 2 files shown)


ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
PR updated with a new commit

ElaadF:
PR updated with a new commit

change-validation/src/main/scala/bootstrap/rudder/plugin/ChangeValidationConf.scala
```diff
@@ -134,13 +135,32 @@ class ChangeValidationWorkflowLevelService(
    * Methode to use to combine several validationNeeded check. It's the same for all objects?
```
ncharles, line 135: you should probably add a comment there saying that it always skips workflow for validated user

ElaadF:
PR updated with a new commit

change-validation/src/main/scala/bootstrap/rudder/plugin/ChangeValidationConf.scala
```diff
@@ -131,16 +132,36 @@ class ChangeValidationWorkflowLevelService(
...
+    /*
+     * Here we check if there is a validated user that should not be subject to any validation workflow
+     * if there is no validated user, we iterate over `checks: Seq[ValidationNeeded]` to verify if there
+     * there is a specific workflow.
+     *
+     * Check why we decided to separate the validated user logic from `ValidationNeeded` objects :
+     * https://issues.rudder.io/issues/22188#note-5
+     */
+    validatedUserRepo.get(actor) match {
+      case Full(ea) =>
```
VinceMacBuche, line 158: You should at least log the error in the emptybox case and i do even think it should be an error

Also you can do both match in one go

validatedUserRepo.get(actor) match {
  case Full(Some(_)
  case Full(None)
  case eb : Emptybox => eb  (i think it should be an error (or ar least log here then getWorkflowAux, but I think it's an error case)

ElaadF:
PR updated with a new commit

VinceMacBuche:
The bug here was that UserValidationNeeded was not  used.

ElaadF:
OK, squash merging this PR

PR #536, opened 2023-02-08
Fixes #22349: Update user plugin to manage update custom roles

fanf:
https://issues.rudder.io/issues/22349

There is a lot of things to change in the way that plugin works, but in that PR I limited to what is the minimum to make it compatible with Rudder 7.3 changes + an auto-backup of `rudder-user.xml` file in case of problem - because we didn't have anything for that, and it would make rudder broken. 

So, the changes related to API changes in core are: 
- reflect name changes from rudder (`RudderRoles`)
- update function from `Box` to `ZIO` (reflecting changes in backend + general direction of where we are going)

Other changes that were needed:
- add a backup/rollback logic if something goes badly when writting the new `rudder-users.xml` file. At least, let the user know that he is not hopeless. 
- remove useless duplicate call to reload after each changes. The reload is done in back-end (no the responsibility of UI). I changed to call to `getUserList` to force refresh UI, which seemed to be the wanted logic, but even that should be removed: the API return of POST should send the wanted updated model, and UI just handle that, which will remove more useless roundtrip to backend.

What is not done but need to be done to have a "user management plugin with custom roles":
- bare minimum: rework the role coverture logic to keep custom role. If the user is defined with some role, the plugin MUST NOT change them, it's only when the role are defined from authz from the UI
- API need to be changed to use ZIO and our API tooling, which implies also changing JSON serialisation
- add the possibility to define custom role if the extended authz plugin is there
- a correct handling of errors from the UI side. We just get a blanck screen with a "error happened" notification, without any way for the user to know what to do. At least say "try to reload", or better handle the error correctly to let the user know things went wrongly, but we did what was needed which is XXX.

### DIFF
--- a/user-management/src/main/elm/sources/UserManagement.elm
+++ b/user-management/src/main/elm/sources/UserManagement.elm
@@ -105,7 +105,7 @@ update msg model =
         AddUser result ->
              case result of
                  Ok username ->
-                     (model, postReloadConf model )
+                     (model, getUsersConf model )
                          |> createSuccessNotification (username ++ " have been added")
                  Err err ->
                      processApiError err model
@@ -113,7 +113,7 @@ update msg model =
         DeleteUser result ->
              case result of
                   Ok deletedUser ->
-                       ({model | panelMode = Closed, login = "", openDeleteModal = False}, postReloadConf model)
+                       ({model | panelMode = Closed, login = "", openDeleteModal = False}, getUsersConf model)
                            |> createSuccessNotification (deletedUser ++ " have been deleted")
                   Err err ->
                        processApiError err model
@@ -121,7 +121,7 @@ update msg model =
         UpdateUser result ->
              case result of
                   Ok username ->
-                      (model, postReloadConf model)
+                      (model, getUsersConf model)
                         |> createSuccessNotification (username ++ " have been modified")
 
                   Err err ->
--- a/user-management/src/main/scala/com/normation/plugins/usermanagement/DataTypes.scala
+++ b/user-management/src/main/scala/com/normation/plugins/usermanagement/DataTypes.scala
@@ -39,7 +39,7 @@ package com.normation.plugins.usermanagement
 
 import bootstrap.liftweb.PasswordEncoder
 import bootstrap.liftweb.RudderConfig
-import bootstrap.liftweb.UserDetailList
+import bootstrap.liftweb.ValidatedUserList
 import com.normation.rudder.Role
 import com.normation.rudder.Role.Custom
 import net.liftweb.common.Logger
@@ -56,7 +56,7 @@ object UserManagementLogger extends Logger {
 
 object Serialisation {
 
-  implicit class AuthConfigSer(auth: UserDetailList) {
+  implicit class AuthConfigSer(auth: ValidatedUserList) {
     def toJson: JValue = {
       val encoder: String = PassEncoderToString(auth)
       val authBackendsProvider = RudderConfig.authenticationProviders.getConfiguredProviders().map(_.name).toSet
@@ -82,7 +82,7 @@ object Serialisation {
     }
   }
 
-  def PassEncoderToString(auth: UserDetailList): String = {
+  def PassEncoderToString(auth: ValidatedUserList): String = {
     auth.encoder match {
       case PasswordEncoder.MD5    => "MD5"
       case PasswordEncoder.SHA1   => "SHA-1"
--- a/user-management/src/main/scala/com/normation/plugins/usermanagement/DataTypes.scala
+++ b/user-management/src/main/scala/com/normation/plugins/usermanagement/DataTypes.scala
@@ -40,12 +40,14 @@ package com.normation.plugins.usermanagement
 import bootstrap.liftweb.PasswordEncoder
 import bootstrap.liftweb.RudderConfig
 import bootstrap.liftweb.ValidatedUserList
-import com.normation.rudder.Role
 import com.normation.rudder.Role.Custom
+import com.normation.rudder.RudderRoles
+
 import net.liftweb.common.Logger
 import net.liftweb.json.{Serialization => S}
 import net.liftweb.json.JsonAST.JValue
 import org.slf4j.LoggerFactory
+import com.normation.zio._
 
 /**
  * Applicative log of interest for Rudder ops.
@@ -64,7 +66,7 @@ object Serialisation {
       val jUser            = auth.users.map {
         case (_, u) =>
           val (rs, custom) = {
-            UserManagementService.computeRoleCoverage(Role.values, u.authz.authorizationTypes).getOrElse(Set.empty).partition {
+            UserManagementService.computeRoleCoverage(RudderRoles.getAllRoles.runNow.values.toSet, u.authz.authorizationTypes).getOrElse(Set.empty).partition {
               case Custom(_) => false
               case _         => true
             }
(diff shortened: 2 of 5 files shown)


user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
```diff
@@ -164,108 +193,96 @@ object UserManagementService {
...
-              Failure(s"Wrong formatting : ${file.path}")
-          }
-        }
-    }
+  def getAll: IOResult[UserFileInfo] = {
+    for {
+      file       <- getUserFilePath
+      parsedFile <- IOResult.attempt(ConstructingParser.fromFile(file.toJava, preserveWS = true))
+      userXML    <- IOResult.attempt(parsedFile.document().children)
+      toUpdate    = (userXML \\ "authentication").head
```
VinceMacBuche, line 270: I don't think this is needed

user-management/src/main/elm/sources/UserManagement.elm
```diff
@@ -105,23 +105,23 @@ update msg model =
         AddUser result ->
              case result of
                  Ok username ->
-                     (model, postReloadConf model )
+                     (model, getUsersConf model )
```
fanf, line 108: we don't need to ask for a reload, it's the role of the backend to ensure backend logic. In fact, the the API return value should be enough to update the UI as wanted.

user-management/src/main/scala/com/normation/plugins/usermanagement/DataTypes.scala
```diff
@@ -56,15 +58,15 @@ object UserManagementLogger extends Logger {
 
 object Serialisation {
 
-  implicit class AuthConfigSer(auth: UserDetailList) {
+  implicit class AuthConfigSer(auth: ValidatedUserList) {
```
fanf, line 61: changes in rudder core

user-management/src/main/scala/com/normation/plugins/usermanagement/DataTypes.scala
```diff
@@ -56,15 +58,15 @@ object UserManagementLogger extends Logger {
...
     def toJson: JValue = {
       val encoder: String = PassEncoderToString(auth)
       val authBackendsProvider = RudderConfig.authenticationProviders.getConfiguredProviders().map(_.name).toSet
 
       val jUser            = auth.users.map {
         case (_, u) =>
           val (rs, custom) = {
-            UserManagementService.computeRoleCoverage(Role.values, u.authz.authorizationTypes).getOrElse(Set.empty).partition {
+            UserManagementService.computeRoleCoverage(RudderRoles.getAllRoles.runNow.values.toSet, u.authz.authorizationTypes).getOrElse(Set.empty).partition {
```
fanf, line 69: need to ask for all roles including custom roles now

user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
```diff
@@ -102,52 +133,48 @@ object UserManagementService {
 
   def computeRoleCoverage(roles: Set[Role], authzs: Set[AuthorizationType]): Option[Set[Role]] = {
 
-    def compareRights(r: Role): Option[Role] = {
-      if (r.name == "no_rights") {
+    def compareRights(r: Role, as: Set[AuthorizationType]): Option[Role] = {
+      if (r == Role.NoRights) {
```
fanf, line 137: we have type, use them. We should not revert to string comparison appart for very good cases (check at system boudaries: parsing, serialisation, etc)

user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
```diff
@@ -102,52 +133,48 @@ object UserManagementService {
...
+            Some(Custom(new Rights(cr.toSeq: _*)))
+          case _                                       => None
         }
       }
     }
 
-    if (authzs.isEmpty || roles.isEmpty || authzs.exists(_.id == "no_rights")) {
+    if (authzs.isEmpty || roles.isEmpty || authzs.contains(AuthorizationType.NoRights)) {
       None
+    } else if (authzs.contains(AuthorizationType.AnyRights)) {
```
fanf, line 153: missing that case, which is now necessary given changes in rudder core

user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
fanf: :+1:

fanf:
PR updated with a new commit

user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
```diff
@@ -49,41 +52,69 @@ object UserOrigin {
...
+                    }
+      } yield ()
     }
   }
 
-  def getUserFilePath: Either[UserConfigFileError, File] = {
-    val resources: Either[UserConfigFileError, UserFile] = UserFileProcessing.getUserResourceFile()
-    resources match {
-      case Left(err) =>
-        Left(err)
```
fanf, line 78: when you see that you write `case Left(err) => Left(err)`, you want to just `map`

user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
```diff
@@ -102,52 +133,48 @@ object UserManagementService {
...
-        val authzNames     = authzs.map(_.id)
-        val roleAuthzNames = r.rights.authorizationTypes.map(_.id)
-        val commonRights   = roleAuthzNames.intersect(authzNames)
+        val commonRights = r.rights.authorizationTypes.intersect(as)
         commonRights match {
           // Intersection is total
-          case cr if cr == roleAuthzNames => Some(r)
-          case cr if cr.nonEmpty          =>
-            val test = cr.flatMap(RoleToRights.parseAuthz)
-            if (test.isEmpty) {
```
fanf, line 117: test can't be empty, there is a check on that in the pattern match (not exactly, but there is no need to parse either, we just have to avoid to `map(_.id)` and remains with the types)

user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
```diff
@@ -102,52 +133,48 @@ object UserManagementService {
...
+          case cr if cr == r.rights.authorizationTypes => Some(r)
+          case cr if cr.nonEmpty                       =>
+            Some(Custom(new Rights(cr.toSeq: _*)))
+          case _                                       => None
         }
       }
     }
 
-    if (authzs.isEmpty || roles.isEmpty || authzs.exists(_.id == "no_rights")) {
+    if (authzs.isEmpty || roles.isEmpty || authzs.contains(AuthorizationType.NoRights)) {
```
fanf, line 151: keep the types! More info, more efficient, avoid useless parse/serialise roundtrip with errors to manage

fanf:
PR updated with a new commit

user-management/src/test/scala/com/normation/plugins/usermanagement/RoleComputationTest.scala
```diff
@@ -4,165 +4,80 @@ import com.normation.plugins.usermanagement.UserManagementService.computeRoleCov
...
-        case Some(r) =>
-          r.head match {
-            case Custom(_) => success
-            case knowRole  => failure(s"Unexpected role : ${knowRole.name}")
-          }
-        case None    => success
-      }
+      computeRoleCoverage(Set(Role.User), Set(AuthorizationType.Compliance.Read)) must beEqualTo(
+        Some(Set(Role.forAuthz(AuthorizationType.Compliance.Read)))
+      )
```
fanf, line 33: Test need to be kept as small as possible, else the intent is lost.  There is a lot of builtin comparators, use them

VinceMacBuche (approved):

PR #546, opened 2023-03-21
Fixes #22501: API selft-service token is not in user management but in API extended authz

fanf:
https://issues.rudder.io/issues/22501


amousset (approved):

fanf:
OK, merging this PR

PR #547, opened 2023-03-21
Fixes #14330: By default, no group are supervised and new group are not supervised 

fanf:
https://issues.rudder.io/issues/14330

Make the all groups being supervised by default, and new groups automatically supervised. 
To achieve that without complexity and a need to look to group creation, we just invert what we save in base: we save unsupervised target, and invert the list to get the supervised ones (we were already retrieving all the groups where performance matters, and the invert is trivial, so it should not change much regarding perf). 

To make migration transparent, on first use we read all supervised targets, invert set based on current group lib, and save the new set of unsupervised targets.
We change the name of the file were supervised groups were persisted to `unsupervised-groups.json`, which give us a way to know if the migration already happened.

We still need to keep the old "supervised" json extractor because API still send supervised targets. I prefered to change as little things as possible.

This PR also harmonize around `supervised` vs the very old `monitored`

### DIFF
--- a/change-validation/src/main/scala/bootstrap/rudder/plugin/MigrateSupervisedGroups.scala
+++ b/change-validation/src/main/scala/bootstrap/rudder/plugin/MigrateSupervisedGroups.scala
@@ -1,0 +1,111 @@
+/*
+*************************************************************************************
+* Copyright 2023 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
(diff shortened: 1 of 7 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR rebased

change-validation/src/main/scala/bootstrap/rudder/plugin/MigrateSupervisedGroups.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  def migrate(): Unit = {
+    Box.tryo {
+      val path = directory + "/" + oldFilename
+      val old = File(path)
+      if (!old.exists) { // ok, plugin installed in new version
+        ChangeValidationLogger.debug("No migration needed from supervised to unsupervised groups")
+      } else { // migration needed
+        ChangeValidationLogger.info(s"Old file format for supervised target found: '${path}': migrating")
+        val oldTargetStrings = net.liftweb.json.Serialization.read[OldfileFormat](old.contentAsString(StandardCharsets.UTF_8))
```
VinceMacBuche, line 76: You could have used zio.json no ?

VinceMacBuche (approved):

VinceMacBuche:
OK, merging this PR

PR #550, opened 2023-03-28
Fixes #12752: Datasources plugin scheduler problems after updating/deleting a datasource

fanf:
https://issues.rudder.io/issues/12752

The problem was an non evaluated `RudderIO` on save :/
I've added a test to check for that now.

### DIFF
--- a/datasources/src/main/scala/com/normation/plugins/datasources/Repository.scala
+++ b/datasources/src/main/scala/com/normation/plugins/datasources/Repository.scala
@@ -168,30 +168,30 @@ class DataSourceRepoImpl(
   /*
    * Be careful, ALL modification to datasource must be synchronized
    */
-  private[this] object datasources extends AnyRef {
-    private[this] val semaphore   = Semaphore.make(1).runNow
-    private[this] val internalRef = Ref.make(Map[DataSourceId, DataSourceScheduler]()).runNow
+  private[datasources] object datasources extends AnyRef {
+    private[this] val semaphore          = Semaphore.make(1).runNow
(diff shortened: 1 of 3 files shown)


datasources/src/main/scala/com/normation/plugins/datasources/Repository.scala
```diff
@@ -168,30 +168,30 @@ class DataSourceRepoImpl(
...
+    private[datasources] val internalRef = Ref.make(Map[DataSourceId, DataSourceScheduler]()).runNow
 
     // utility methods on datasources
     // stop a datasource - must be called when the datasource still in "datasources"
-    private[this] def stop(id: DataSourceId) = {
+    private[datasources] def stop(id: DataSourceId) = {
       DataSourceLoggerPure.debug(s"Stopping data source with id '${id.value}'") *>
-      internalRef.get.map(_.get(id) match {
-        case None      => DataSourceLogger.trace(s"Data source with id ${id.value} was not found running")
+      internalRef.get.flatMap(_.get(id) match {
```
fanf, line 179: this is the problem. `dss.cancel()` was never evaluated. It dates from rudder 6.0 !

VinceMacBuche: It made an IOResult [IOResult[Unit]] ?

VinceMacBuche: Maybe we could add return type of stop method

fanf:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #554, opened 2023-04-05
Fixes #22601: Missing validation button with admin

fanf:
https://issues.rudder.io/issues/22601


change-validation/src/main/scala/com/normation/plugins/changevalidation/snippet/ChangeRequestDetails.scala
```diff
@@ -167,12 +170,20 @@ class ChangeRequestDetails extends DispatchSnippet with Loggable {
   }
 
   def displayActionButton(cr: ChangeRequest, step: WorkflowNodeId): NodeSeq = {
-    val authz   = CurrentUser.getRights.authorizationTypes.toSeq.collect { case right: ActionType.Edit => right.authzKind }
+    val authz   = Nil // we are sideStepping it, see: https://issues.rudder.io/issues/22595
```
fanf, line 173: this is the problem: we don't look for `any` because we are comaring on strings, not with the correct `Authorization` method. 
We can change that now, see: https://issues.rudder.io/issues/22595

change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
```diff
@@ -99,22 +102,39 @@ class EitherWorkflowService(cond: () => Box[Boolean], whenTrue: WorkflowService,
...
+   * Validation rule logic:
+   * - check for self validation (only needed if current user is the CR author)
+   * - check for current user rights.
+   *   WARNING: WE ARE SIDE STEPPING authz check until https://issues.rudder.io/issues/22595 is solved
+   */
+  private def canValidate(isCreator: Boolean, selfValidation: () => Box[Boolean]): Boolean = {
+    val correctActor = selfValidation().getOrElse(false) || !isCreator
+    correctActor && CurrentUser.checkRights(AuthorizationType.Validator.Edit)
+  }
+
```
fanf, line 132: These two methods are the relevant changes: we use `CurrentUser.checkRights` in place of a comparison on string. That methods correctly check for the `any` right that admin has.

VinceMacBuche:
We can't remove currentUserRights: Seq[String] parameters because they are defined in the trait within Rudder ?

VinceMacBuche (approved):

PR #557, opened 2023-05-09
Fixes #22738: OIDC provided custom role list

fanf:
https://issues.rudder.io/issues/22738

We already have a class in charge of creating a "rudder user" from a "oidc token" (`RudderOidcUserService`). So that PR only customize the "load roles" part so that in addition to the rudder roles, we look at the added configuration parameters of the OIDC providers to see if and what we have to do with OIDC list of role name provided in the token. 

The role resolution from name use the existing logic for the role names in `rudder-users.xml`, nothing new here. 

The most complicated part was the configuration properties part (bring the 3 new property values to the RudderOidcUserService). It needed changes in the spring beans declaration, and it's always a mess. 

Most of the other changes are doc and logging things so that it is very clear that an OIDC user uses roles from [`rudder-users.xml`|OIDC token|both].

### DIFF
--- a/auth-backends/README.adoc
+++ b/auth-backends/README.adoc
@@ -13,13 +13,18 @@ other general information.
 = Authentication backends
 
 This plugins allows to use alternative authentication backends for Rudder: *SAMLv2*, *OpenID Connect*, and *LDAP or Active Directory (AD)*.
-The old radius module is deprecated and will totally removed in a future version.
+The old *radius module is deprecated* and will be totally removed in a future version.
 
-Each authentication method is detailed below. Users are expected to know how an authentication system works independently from Rudder to configure it in Rudder: you will likely need authentication token, URLs, and other properties provided by your company.
+Each authentication method is detailed below. Users are expected to know how an authentication system works independently of Rudder to configure it in Rudder: you will likely need authentication token, URLs, and other properties provided by your company.
+
+== Centralized user authorization management with OIDC
+
+In addition to alternative authentication provider, the OIDC backends allow to provide Rudder roles through the OIDC token so that you can control your Rudder user authorisation directly from your identity provider.
+See OIDC chapter below for more information.
 
 == Configure login form rendering
 
-By default, the standard Rudder login form is displayed. SSO backends like OpenID Connect ones add links to the relevant SSO own login page below that Rudder login form. When you use such an authentication method, you want to hide or totaly remove Rudder own login form to avoid to confuse you user. For that, you can use the following property:
+By default, the standard Rudder login form is displayed. SSO backends like OpenID Connect ones add links to the relevant SSO own login page below that Rudder login form. When you use such an authentication method, you want to hide or totally remove Rudder own login form to avoid to confuse you user. For that, you can use the following property:
 
 ```
 rudder.auth.displayLoginForm=show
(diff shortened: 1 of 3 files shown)


fanf:
PR updated with a new commit

auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
```diff
@@ -484,13 +499,68 @@ trait RudderUserServerMapping[R <: OAuth2UserRequest, U <: OAuth2User, T <: Rudd
...
+        registrationRepository.registrations.get(userRequest.getClientRegistration.getRegistrationId) match {
+          case None      =>
+            AuthBackendsLogger.trace(
+              s"No configuration found for OIDC registration id: ${userRequest.getClientRegistration.getRegistrationId}"
+            )
+            rudder.roles // if no registration, use user roles
+          case Some(reg) =>
+            if (reg.roles.enabled) {
+              val custom = {
+                try {
```
fanf, line 521: here is the main part that hav impact on role, ie the security aspect. 
- we use standard oidc data type, 
- findRoleByName only looks for existing role with the same name (case insensitive, like other role resolution in rudder)

auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
```diff
@@ -484,13 +499,68 @@ trait RudderUserServerMapping[R <: OAuth2UserRequest, U <: OAuth2User, T <: Rudd
...
+
+              if (custom.nonEmpty) {
+                ApplicationLoggerPure.Authz.logEffect.info(
+                  s"Principal '${rudder.getUsername}' role list extended with OIDC provided roles: [${custom.toList.map(_.name).sorted.mkString(", ")}] (override: ${reg.roles.over})"
+                )
+              }
+
+              val roles = if (reg.roles.over) {
+                // override means: don't use user role configured in rudder-users.xml
+                custom
```
fanf, line 546: perhaps we still need to give the `user` role which is added automatically to every users, else it might breaks some assumption. But on the other hand, adding it automatically breaks the assumption "override means that you get only the role in the list provided by the idP"

auth-backends/README.adoc
```diff
@@ -13,13 +13,18 @@ other general information.
 = Authentication backends
 
 This plugins allows to use alternative authentication backends for Rudder: *SAMLv2*, *OpenID Connect*, and *LDAP or Active Directory (AD)*.
-The old radius module is deprecated and will totally removed in a future version.
+The old *radius module is deprecated* and will totally be removed in a future version.
```
amousset, line 16:
```suggestion
The old *radius module is deprecated* and will be totally removed in a future version.
```

auth-backends/README.adoc
```diff
@@ -543,7 +586,7 @@ In the log, you see:
 It means that the value used for `rudder.auth.oauth2.provider.${registrationKey}.userNameAttributeName` was correctly returned in the profile list for the authenticated user, but that value was not found in Rudder user configuration files `/opt/rudder/etc/rudder-users.xml`.
 Check that one of the entries in that file has the corresponding value for its `name` attribute.
 
-=== Radius backend
+=== Radius backend (deprecated)
```
fanf, line 589: "deprecated" is already in a warn info notice just below, but some people were still missing it - nothing to do with the main pr, it's just cleaning things up like other typos.

VinceMacBuche (approved):

PR #559, opened 2023-05-12
Fixes #22757: makes explicit that the plugin is discountinued

fanf:
https://issues.rudder.io/issues/22757


fanf:
OK, merging this PR

PR #562, opened 2023-05-15
Use FQDN in FusionInventories

BeowolfK:
Modify short hostname used by fusion inventories by FQDN used in Rudder


BeowolfK:
According to PR #573 , add CLI option for FQDN to not change global conf and may able to just send-one with FQDN

PR #574, opened 2023-06-28
Fixes #22967: Spurious test error about length of generated password

fanf:
https://issues.rudder.io/issues/22967

So the problem seems to be that depending of the timing, when we observe `f1` status, it is either still running because init if not done (typically on a slow machine where the foreign datasource answer takes time to be completed) or suspended. 

Now, the test check for both status (they are both ok at that point). 
I also added a new check that the reloaded datasource fiber is actually live.


fanf:
Bad ticket for PR, closing

PR #577, opened 2023-07-10
Fixes #23047: Remove radius support in Rudder 8.0

fanf:
https://issues.rudder.io/issues/23047

Delete radius and all dependencies/occurence, including doc, elm, etc. Nothing special appart a big relief :)


fanf:
I don't understand the reported problems by CI for that one. I will merge by hand and check if things are broken then

PR #581, opened 2023-07-21
Fixes #23112: Add the role-retrieving OIDC feature to OAuth2

fanf:
https://issues.rudder.io/issues/23112

The role mapping was only done in a OIDC case, in a ` buildUser` function in the OIDC specialized `loadUser` function.
That PR factorize-out the `buildUser` function to put it into the generic `RudderUserServerMapping` trait. That works well because we retrieve group attribute with a function (`getAttribute[java.util.ArrayList[String]](reg.roles.attributeName)`) that is common to both protocol. 
The remaining bits are just gluing:
- pass the protocol name as a parameter for have correct logs
- add the `RudderClientRegistrationRepository` in the trait because it is now required for both, and change class parameters to match it
- generalize the user details creation to adapt to both `OAuth2` and `OIDC` case in there respective implementation.

Also correct a pair of unrelated typos.

### DIFF
--- a/auth-backends/src/main/scala/com/normation/plugins/authbackends/AuthBackendsRepository.scala
+++ b/auth-backends/src/main/scala/com/normation/plugins/authbackends/AuthBackendsRepository.scala
@@ -102,7 +102,7 @@ class AuthBackendsRepository(
 
     val file = JsonFileConfig(
       "file",
-      """By default, Rudder authentication is collocated with user authorisation
+      """By default, Rudder authentication is collocated with user authorization
         |in 'rudder-user.xml' file. The file is located:""".stripMargin,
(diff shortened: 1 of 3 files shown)


auth-backends/src/main/scala/com/normation/plugins/authbackends/AuthBackendsRepository.scala
```diff
@@ -102,7 +102,7 @@ class AuthBackendsRepository(
 
     val file = JsonFileConfig(
       "file",
-      """By default, Rudder authentication is collocated with user authorisatoin
+      """By default, Rudder authentication is collocated with user authorisation
```
amousset, line 105:
```suggestion
      """By default, Rudder authentication is collocated with user authorization
```

PR #582, opened 2023-07-27
Fixes #23141: Schedule not displayed correctly in logs

amousset:
https://issues.rudder.io/issues/23141


fanf (approved):
Doesn't seems that there is a way to display a schedule anymore in ZIO :/

PR #583, opened 2023-07-27
Fixes #22993: Can't deploy, “Configuration state of the object modified by this change request has changed” looping

ElaadF:
https://issues.rudder.io/issues/22993

### What is the bug
The change-request was blocked in state "deployment" but can be deployed (the changes was deployed even if the change-request is still in the "pending deployment" section. The problem only occurs for pending deployment request, the step valaidation to deployed, is working as expected.

### Why it happens

The bug was that it was missing the logic for deployment to deployed part
The bug is only present in 7.3 due to https://github.com/Normation/rudder-plugins/commit/3103cf4338d98ed30425e4fbb5e31ba449433141 the logic was removed 

### How can I validate that the fix is working ?

1. install the plugin change-validation
2. create a group and add node in it
3. activate change request in the plugin page
4. modify the name
5. Check that a change request is generated
6. Validate the change request, the change request should be in "Pending deployment"
7. Deploy the change request, it should be deployed and no longer in the "Pending deployment" page
8. Try the same thing but from validation to deployed directly

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
@@ -208,6 +208,7 @@ class TwoValidationStepsWorkflowServiceImpl(
       currentStep:       WorkflowNodeId,
       isCreator:         Boolean
   ): WorkflowAction = {
+
     val deployAction = {
       if (canDeploy(isCreator, selfDeployment))
         Seq((Deployed.id, stepValidationToDeployed _))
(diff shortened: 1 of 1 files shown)


change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
```diff
@@ -435,6 +441,15 @@ class TwoValidationStepsWorkflowServiceImpl(
     toFailure(Deployment, changeRequestId, actor, reason)
   }
 
+  private[this] def stepDeploymentToDeployed(
```
ElaadF, line 444: This was removed in 7.3 by https://github.com/Normation/rudder-plugins/commit/3103cf4338d98ed30425e4fbb5e31ba449433141#diff-98e8dc3343e38546c42491a39303758ad01214ed8e5ef40c6a2c4d85cfa73f8eL413

This is a missing steps, don't understand why it's have been removed

change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
```diff
@@ -224,7 +225,12 @@ class TwoValidationStepsWorkflowServiceImpl(
         WorkflowAction("Validate", validatorActions)
 
       case Deployment.id     =>
-        WorkflowAction("Deploy", deployAction)
+        val deploymentAction = {
```
ElaadF, line 228: https://github.com/Normation/rudder-plugins/commit/3103cf4338d98ed30425e4fbb5e31ba449433141#diff-98e8dc3343e38546c42491a39303758ad01214ed8e5ef40c6a2c4d85cfa73f8eL205 if think this is a oversight

Without this we cannot close the change-request, the change-request will be deployed, be will still remains in the deployment step

ncharles (approved):

PR #592, opened 2023-08-18
Fixes #23254: User management plugin incorrectly understand OIDC roles

fanf:
https://issues.rudder.io/issues/23254

A first PR for the problematic of dynamic role. This one is done so that it is compatible with current Rudder 7.3 and so we don't have the real configuration option for OIDC role override. So we just add a generic warning message to avoid surprise. 

This PR also correct the fact that the list of authentication providers is hard coded in user management (to provide human readable version ?) and that authentication provider out of that not-maintained list are IGNORED. 
At some point, we will have a type for providers, with their properties (does it override roles, etc), but we should never ignore unknown providers! I just removed the types, because they would likely be something else. 
This was why OIDC was not listed and so user got false information.

Message:
image
image

### DIFF
--- a/user-management/README.adoc
+++ b/user-management/README.adoc
@@ -16,7 +16,7 @@ This plugin allows to get new capabilities for user management:
 - and a UI is added in Rudder `Administration` menu to help with user management.
 
 Note that the xref:plugins:api-authorizations.adoc[API authorizations] plugin also extends user
-rights with a self-service UI to manage an API token to perform API calls with their personal account. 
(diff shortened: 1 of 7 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #596, opened 2023-08-24
Fixes #23348: not allowed to access errors because rudder plugins are missing AuthorizationApiMapping 

fanf:
https://issues.rudder.io/issues/23348

Use the new facility provided by https://github.com/Normation/rudder/pull/5004 to declare rights along with the API schema. Correct a bunch of API rights which were using the default one (`Administrator.Write`) in place of something less powerful (often `Administration.Read`)

For change validation: 
- the dedicated api right mapper is not useful anymore, 
- it was missing `ValidatedUser` in the list of aggregated endpoints, which was causing specific additional problems.


fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #602, opened 2023-10-12
Fixes #23563: When a datasource is deleted, its properties should be deleted

fanf:
https://issues.rudder.io/issues/23563

Delete property related to datasource when the datasource is deleted. 

Main changes: 
- factor out the logic to update the node property for the datasource so that it can be used in both update and delete method, 
- add a `cause` to delete. Likely other methods from the repo should have it too. 
- add a `deleteAll` method in `HttpQueryDataSourceService` because it's really it that has the logic to things on node, not the repository (the repo is just the warrant for datasource backend/liveness consistency)
- add a unit test that checks that the properties are really gone when the datasource is deleted.

Note: deletion of a datasource will only remove properties with the `datasource` provider. 
It allows the following use case: 
- configure datasources with "on missing (404), keep node properties"
- get values from remote provider, but for the missing one, set value by hand one nodes
- if you delete the datasource, the properties manage by hand are still there, managed by hand. 

If you don't want to delete properties, you can just disable the datasource in place of deleting it. 
A latter feature could be to add a check box in the deletion confirmation pop-up, asking for what to do.

### DIFF
--- a/datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
+++ b/datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
@@ -45,6 +45,8 @@ import com.normation.rudder.domain.nodes.NodeState
 import com.normation.rudder.domain.policies.GlobalPolicyMode
 import com.normation.rudder.domain.properties.CompareProperties
 import com.normation.rudder.domain.properties.GlobalParameter
+import com.normation.rudder.domain.properties.NodeProperty
+import com.normation.rudder.domain.properties.PropertyProvider
 import com.normation.rudder.repository.RoParameterRepository
 import com.normation.rudder.repository.WoNodeRepository
 import com.normation.rudder.services.nodes.NodeInfoService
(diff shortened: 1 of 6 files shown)


fanf:
We must take care to only delete properties that were set by the datasources plugin with that name.

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
```diff
@@ -83,6 +85,54 @@ trait QueryDataSourceService {
...
+      optProperty:       Option[NodeProperty],
+      cause:             UpdateCause,
+      enforceSameOrigin: Boolean // for deletion, we want to be able to remove only prop set by datasource provider
+  ): IOResult[NodeUpdateResult] = {
+    optProperty match {
+      // on none, don't update anything, the life is wonderful (because 'none' means 'don't update')
+      case None           => NodeUpdateResult.Unchanged(nodeInfo.id).succeed
+      case Some(property) =>
+        // look for the property value in the node to know if an update is needed.
+        // we only care about value here (not provider or other meta-info)
```
clarktsiory, line 113: I believe this comment is no longer up-to-date with the new behavior : it looks like we also care about the provider when `enforceSameOrigin` is true

clarktsiory (commented):
Apart from the outdated comment, everything looks good to me !

datasources/src/main/scala/com/normation/plugins/datasources/Repository.scala
```diff
@@ -297,10 +298,16 @@ class DataSourceRepoImpl(
   /*
    * delete need to clean existing live resource
    */
-  override def delete(id: DataSourceId): IOResult[DataSourceId] = dataSourcesLock.withPermit {
+  override def delete(id: DataSourceId, cause: UpdateCause): IOResult[DataSourceId] = dataSourcesLock.withPermit {
     // start by cleaning
     datasources.delete(id) *>
-    backend.delete(id)
+    backend.delete(id, cause) *>
+    fetch.deleteAll(id, cause) *>
```
VinceMacBuche, line 305: I think we should delete the properties before deleting the datasources, or else if something goes wrong when deleting a property, the datasource is deleted and we can trigger properties deletion again. 

Maybe a clean up job at start up cleaning invalid datasource properties ? ( just thinking out loud)

fanf:
PR updated with a new commit

datasources/src/main/scala/com/normation/plugins/datasources/QueryService.scala
fanf: yes

datasources/src/main/scala/com/normation/plugins/datasources/Repository.scala
fanf: I think it would be more dangerous: we could be in a case where the datasource is there, but some properties related to it would be deleted (so the generation changed while the hypothesis didn't). On that order, at worst we remain in the same situation. 

Yes, for the cleaning, something is a broken right now. A bootstrap check seems a good first solution. I opened that ticket to track it: https://issues.rudder.io/issues/23679

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #603, opened 2023-10-20
Fixes #23444: Impact of adding UserInfo base on public plugins

fanf:
https://issues.rudder.io/issues/23444

That change needs https://github.com/Normation/rudder/pull/5121.

The changes in `user-management` are just the minimum to make things compile. A big change is still needed: we must use information from database in that plugin, which is not done here. 

The changes in `authentication-backend` add the possibility for `oidc/oauth2` backend to auto-provision users in Rudder if their authentication was successful. 
That possibility is governed by a new IdP propery: `rudder.auth.oauth2.provider.${idp}.enableProvisionning=true`

The logic change is then rather small and mainly in [AuthBackendsConf.scala](https://github.com/Normation/rudder-plugins/pull/603/files#diff-8fe302fecad34e77c8d7fa30f3efdf0110c77048e1f14006e4855dc1c60e73d9): on authentication success, we try to retrieve user from user repository. It it is not here and provisioning is allowed, we create it. 

Further evolution can be done to add more details when the user is provisioned, for now only minimal data is created (login, etc).

### DIFF
--- a/auth-backends/README.adoc
+++ b/auth-backends/README.adoc
@@ -475,7 +475,7 @@ so that it is easier to change or disable some of them.
 
 === IdP-provided authorisations for Rudder users
 
-You can configure an OAuth2 or OIDC provider so that it informs Rudder of the roles the user need to have. This allows to centrally
+You can configure an `OAuth2` or `OIDC` provider so that it informs Rudder of the roles the user need to have. This allows to centrally
 manage both user and authorisation in the same place.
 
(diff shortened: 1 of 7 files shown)


fanf:
PR updated with a new commit

fanf:
Since parent was merged, merging for consistency

fanf:
OK, squash merging this PR

PR #607, opened 2023-11-09
Fixes #21724: Categories can't be folded in ACL API with authorization plugin

clarktsiory:
https://issues.rudder.io/issues/21724

Instead of the 'category' title having a padding of 40px, the icon has a width of 40px.
The icon is also in a label now, so it allows to do the fold :partying_face:


clarktsiory:
PR updated with a new commit

clarktsiory:
Commit modified

clarktsiory:
Commit modified

VinceMacBuche (approved):

PR #609, opened 2023-11-20
Fixes #23760: Migrate branding, openscap, scale-out-relay to zio-json

clarktsiory:
https://issues.rudder.io/issues/23760

Removing any occurrence of lift-json in the three plugins : branding, openscap, scale-out-relay.

In the branding plugin there is an outdated serial format (from the v5.0 of rudder), which should be removed later. The openscap plugin has only unused imports from lift-json, I took the opportunity to add yaml tests for the API

### DIFF
--- a/branding/src/main/scala/com/normation/rudder/rest/BrandingApiSchema.scala
+++ b/branding/src/main/scala/com/normation/rudder/rest/BrandingApiSchema.scala
@@ -37,9 +37,8 @@
 
 package com.normation.rudder.rest
 
-import com.normation.rudder.api.HttpAction._
 import com.normation.rudder.AuthorizationType
-
+import com.normation.rudder.api.HttpAction._
 import sourcecode.Line
 
 /*
@@ -55,7 +54,7 @@ object BrandingApiEndpoints extends ApiModuleProvider[BrandingApiSchema] {
     val z              = implicitly[Line].value
     val description    = "Get branding plugin configuration"
     val (action, path) = GET / "branding"
-    val dataContainer: Option[String] = None
+    val dataContainer:  Option[String]          = None
     override def authz: List[AuthorizationType] = List(AuthorizationType.Administration.Read)
   }
 
--- a/branding/src/test/scala/com/normation/plugins/branding/api/BrandingApiTest.scala
+++ b/branding/src/test/scala/com/normation/plugins/branding/api/BrandingApiTest.scala
@@ -1,0 +1,87 @@
+package com.normation.plugins.branding
+
+import better.files._
+import com.normation.plugins.branding.api.BrandingApi
+import com.normation.plugins.branding.api.BrandingApiService
+import com.normation.rudder.api.ApiVersion
+import com.normation.rudder.rest.RestTestSetUp
+import com.normation.rudder.rest.TraitTestApiFromYamlFiles
+import java.nio.file.Files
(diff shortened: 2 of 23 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

branding/src/main/scala/com/normation/rudder/rest/BrandingApiSchema.scala
```diff
@@ -64,13 +63,15 @@ object BrandingApiEndpoints extends ApiModuleProvider[BrandingApiSchema] {
     val description    = "Update branding plugin configuration"
     val (action, path) = POST / "branding"
     val dataContainer: Option[String] = None
+    override val name: String         = "updateBRandingConf" // the initial version of the plugin used that name
```
VinceMacBuche, line 66: Indeed but i think we can fix it. It's an api from plugin that is not used much, especially from api and we better not keep it 

That's good rule to keep it in most cases but this one is really not dangerous ... and i can live with that typo if we decide to keep it

clarktsiory: Okay, I did not take the risk because I didn't know the use case, but I'm changing it then :rocket: 
 
Also, there is the endpoint `ReloadBrandingConf` that currently has the same name as `GetBrandingConf`, I guess I can change it on the same basis to have a name _reloadBrandingConf_

clarktsiory:
PR updated with a new commit

branding/src/main/scala/com/normation/rudder/rest/BrandingApiSchema.scala
VinceMacBuche: Yes totally !

VinceMacBuche: And as I said, not changing the value is in general cases the good move indeed !!

branding/src/test/scala/com/normation/plugins/branding/api/BrandingApiTest.scala
```diff
@@ -0,0 +1,87 @@
...
+import org.specs2.runner.JUnitRunner
+import org.specs2.specification.AfterAll
+
+@RunWith(classOf[JUnitRunner])
+class BrandingApiTest extends Specification with TraitTestApiFromYamlFiles with Loggable with AfterAll {
+  sequential
+
+  val restTestSetUp = RestTestSetUp.newEnv
+
+  val tmpDir: File = File(Files.createTempDirectory("rudder-test-"))
```
fanf, line 22: what we often do is create a directory, and clean it `AfterAll` but only if a jvm parameter is not set, something like: 
```
  override def afterAll(): Unit = {
    if (System.getProperty("tests.clean.tmp") != "false") {
      // do cleanup
    }
  }
```

That allow to keep the information when debugging is needed, you run the test with `-Dtests.clean.tmp=false` and you can explore what is going wrong. 

It's not a mandatory change, and since I think nothing is written by the business code tested here, it's not really interesting (it's just fyi)

fanf (approved):
LGTM, so much cleaning \o/

PR #610, opened 2023-11-20
Fixes #23762: Update rudder-plugins dependencies

VinceMacBuche:
https://issues.rudder.io/issues/23762


fanf (approved):
The password deletion in more recent spring security OAUTH2 should be deal with in an other issue. We don't want to be block because of it since we don't use it

PR #612, opened 2023-11-23
Fixes #23775: Migrate api-authorizations, auth-backends to zio-json

clarktsiory:
https://issues.rudder.io/issues/23775

Migrating api-authorizations, there needed to be a rewrite of [the lift-json serializer](https://github.com/Normation/rudder/blob/master/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/RestDataSerializer.scala#L669) into a data transfer case class and the added tests ensures that the display/hiding of the token value is the same after migrating to zio-json...

### DIFF
--- a/api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/UserTokenApiDefinition.scala
+++ b/api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/UserTokenApiDefinition.scala
@@ -37,31 +37,30 @@
 
 package com.normation.plugins.apiauthorizations
 
-import com.normation.box._
 import com.normation.eventlog.ModificationId
 import com.normation.rudder.api._
-import com.normation.rudder.apidata.ApiAccountSerialisation._
+import com.normation.rudder.apidata.JsonApiAcl
 import com.normation.rudder.rest._
 import com.normation.rudder.rest.{UserApi => API}
+import com.normation.rudder.rest.implicits.ToLiftResponseOne
 import com.normation.rudder.rest.lift._
+import com.normation.utils.DateFormaterService
 import com.normation.utils.StringUuidGenerator
-import net.liftweb.common.EmptyBox
-import net.liftweb.common.Full
+import io.scalaland.chimney.Transformer
 import net.liftweb.http.LiftResponse
 import net.liftweb.http.Req
-import net.liftweb.json._
-import net.liftweb.json.JsonAST.JArray
-import net.liftweb.json.JsonDSL._
 import org.joda.time.DateTime
+import zio.json._
 
 class UserApi(
-    restExtractor:  RestExtractorService,
     readApi:        RoApiAccountRepository,
     writeApi:       WoApiAccountRepository,
     tokenGenerator: TokenGenerator,
     uuidGen:        StringUuidGenerator
 ) extends LiftApiModuleProvider[API] {
   api =>
+
+  import UserApi._
 
   def schemas = API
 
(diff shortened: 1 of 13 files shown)


api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/UserTokenApiDefinition.scala
```diff
@@ -85,34 +84,18 @@ class UserApi(
...
-
-        case eb: EmptyBox =>
-          val e = eb ?~! s"Error when trying to get user '${authzToken.actor.name}' API token"
-          RestUtils.toJsonError(None, e.messageChain)(schema.name, true)
-      }
+      readApi
+        .getById(ApiAccountId(authzToken.actor.name))
+        .map(RestAccountsResponse.fromRedacted(_))
+        .chainError(s"Error when trying to get user '${authzToken.actor.name}' API token")
+        .toLiftResponseOne(params, schema, None)
```
fanf, line 93: the clarity gain is amazing

api-authorizations/src/test/scala/com/normation/plugins/apiauthorizations/api/UserApiTest.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  sequential
+
+  val restTestSetUp = RestTestSetUp.newEnv
+
+  val tmpDir: File = File(Files.createTempDirectory("rudder-test-"))
+  override def yamlSourceDirectory  = "authorizations_api"
+  override def yamlDestTmpDirectory = tmpDir / "templates"
+
+  val fixedDate = DateTime.parse("1970-01-01T00:00:00.000Z")
```
fanf, line 37: Can you use a different, random date in the present ? That specific date and time is too easy to believe as a badly init one ?

api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/UserTokenApiDefinition.scala
```diff
@@ -128,67 +111,172 @@ class UserApi(
...
+      accounts: List[RestApiAccount]
+  )
+
+  object RestAccountsResponse {
+    import RestApiAccount._
+
+    implicit val encoder: JsonEncoder[RestAccountsResponse] = DeriveJsonEncoder.gen[RestAccountsResponse]
+
+    private def apply(accounts: List[ApiAccount], secret: Option[String] = None): RestAccountsResponse = new RestAccountsResponse(
+      accounts.map(a => secret.map(a.toRestWithSecret(_)).getOrElse(a.toRest))
```
fanf, line 237: Can you add a comment explaining that the list is needed b/c the answer expects a json array ?

api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/UserTokenApiDefinition.scala
```diff
@@ -128,67 +111,172 @@ class UserApi(
...
-          val accounts: JValue = ("accounts" -> JArray(Nil))
-          RestUtils.toJsonResponse(None, accounts)(schema.name, true)
+}
+
+object UserApi {
+
+  final case class RestApiAccount(
+      id:                              ApiAccountId,
+      name:                            ApiAccountName,
+      token:                           ApiToken,
```
fanf, line 151: it would make sense to not use `ApiToken`, but a new `ClearTextToken(value: String)` or something, so that it's not the same structure send, and we don't forget to zero it during a refactoring.

api-authorizations/src/test/resources/authorizations_api/api_users.yml
```diff
@@ -0,0 +1,110 @@
...
+    {
+     "action" : "createApiToken",
+     "result" : "success",
+     "data" : {
+       "accounts" : [
+          {
+            "id" : "user1",
+            "name" : "user1",
+            "token" : "generated-test-token",
+            "tokenGenerationDate" : "1970-01-01T01:00:00+01:00",
```
fanf, line 79: try to use different date for different use case, for ex simulating a regenerated token, so that we can see that the correct date ends-up in the correct field.

fanf (commented):
Some little things to change, but that's a super nive evolution in our code base, GG. 
It can be merged once the remark are addressed.

clarktsiory:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #613, opened 2023-11-29
Fixes #23805: Migrate datasource plugin to zio-json

clarktsiory:
https://issues.rudder.io/issues/23805

There are two different use cases of deserialization in the plugin, that were in the `OptionalJson` and `CompleteJson`  extractors : 
 * the _complete_ one for representing a whole datasource properties object with **all fields mandatory**
 * the _optional_ one for representing a subset of them, so with **all fields optional** 

We needed to add case classes with **full** representation of a datasource, and derive json codecs accordingly so that the migration keeps the json format and validation constraints.
The previously named "wrappers" classes, with optional fields has been renamed to **patch** classes, as they are used to update a datasource (patch is the operation name as defined by _chimney_ for that use case). These "patch" classes have the same field names as the "full" classes, the types are only wrapped in `Option[_]`.

Added postgresql tests along the way, and rewrote the existing API tests to the yaml format

### DIFF
--- a/datasources/src/main/scala/com/normation/plugins/datasources/Serialization.scala
+++ b/datasources/src/main/scala/com/normation/plugins/datasources/Serialization.scala
@@ -37,21 +37,28 @@
 
 package com.normation.plugins.datasources
 
-import cats._
+import DataSourceJsonCodec._
 import cats.implicits._
-import com.normation.plugins.datasources.DataSourceSchedule._
-import com.normation.plugins.datasources.HttpRequestMode._
+import com.normation.errors._
 import com.normation.rudder.domain.properties.GenericProperty._
 import com.normation.rudder.repository.json.JsonExtractorUtils
+import com.normation.rudder.rest.RudderJsonRequest._
 import com.typesafe.config.ConfigRenderOptions
+import com.typesafe.config.ConfigValue
+import io.scalaland.chimney.Patcher
+import io.scalaland.chimney.Transformer
+import io.scalaland.chimney.dsl.PatcherConfiguration
+import io.scalaland.chimney.syntax._
 import java.util.concurrent.TimeUnit
-import net.liftweb.common.Box
 import net.liftweb.common.Failure
 import net.liftweb.common.Full
-import net.liftweb.json._
-import net.liftweb.util.ControlHelpers.tryo
+import net.liftweb.http.Req
 import scala.concurrent.duration.FiniteDuration
+import scala.math.BigDecimal
+import scala.util.chaining._
 import zio._
+import zio.json._
+import zio.json.ast.Json
 
 object Translate {
   implicit class DurationToScala(d: Duration) {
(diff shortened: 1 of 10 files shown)


datasources/src/test/scala/com/normation/plugins/datasources/RepositoryTest.scala
```diff
@@ -0,0 +1,127 @@
...
+    // function declaration is an 'update' operation and function call is a 'select' one
+    val (update, select) = sqlText.splitAt(sqlText.indexOf("select create_datasources()"))
+
+    doobie.transactRunEither(xa => (Update0(update, None).run <* Query0(select).option).transact(xa)) match {
+      case Right(_) => ()
+      case Left(ex) => throw ex
+    }
+  }
+
+  val repo = new DataSourceJdbcRepository(doobie)
```
fanf, line 57: you need to guard the init of that class (or others relying on postgres/connection pool) else it will blast at runtime when on a system without a database. Something like that works, even if not great: 
```
  val repo = if(doDatabaseConnection) {
    new DataSourceJdbcRepository(doobie)
  } else null

```

datasources/src/main/scala/com/normation/plugins/datasources/Serialization.scala
```diff
@@ -60,368 +66,521 @@ object Translate {
...
+  ) extends FullDataSourceSchedule
+
+  implicit val transformerFrom: Transformer[FullDataSourceSchedule, DataSourceSchedule] =
+    Transformer.derive[FullDataSourceSchedule, DataSourceSchedule]
+
+  implicit val transformerTo: Transformer[DataSourceSchedule, FullDataSourceSchedule] =
+    Transformer.derive[DataSourceSchedule, FullDataSourceSchedule]
+}
+
+trait DataSourceJsonCodec {
```
fanf, line 309: I'm not sure having that as a trait IF extening traits neet to use the same encoders are a great idea. 
Depending of the order of init, you can get error like that: 
```
java.lang.ExceptionInInitializerError
	at com.normation.plugins.datasources.DataSourceExtractor$PatchDataSourceType$.<clinit>(Serialization.scala:442)
	at com.normation.plugins.datasources.DataSourceJsonCodec.$init$(Serialization.scala:366)
	at com.normation.plugins.datasources.DataSourceJdbcRepository$.<clinit>(Repository.scala:483)
	at com.normation.plugins.datasources.RepositoryTest.$anonfun$new$2(RepositoryTest.scala:92)
	at org.specs2.specification.dsl.mutable.EffectBlocks.tryBlock(EffectBlocks.scala:131)
	at org.specs2.specification.dsl.mutable.EffectBlocks.$anonfun$nestBlock$2(EffectBlocks.scala:110)
        ....
	at com.normation.plugins.datasources.DataSourceExtractor$PatchDataSourceType$.decoder(Serialization.scala:443)
	at com.normation.plugins.datasources.DataSourceJsonCodec.$init$(Serialization.scala:366)
	at com.normation.plugins.datasources.DataSourceExtractor$.<clinit>(Serialization.scala:387)
	... 43 more
```

Here, we have a call to 
```
      check(DataSourceJdbcRepository.insertDataSourceSQL(fakeDataSource.id, fakeDataSource.transformInto[FullDataSource]))
```
which extends `DataSourceJsonCodec`. So we init parent (trait) parameters, ie all the `implicit val` in order. 
We reach the last one, `NewDataSource`, which has a `PatchDataSourceHttpType`. 

So we look for the codec, which is provided by `object PatchDataSourceType` in `object DataSourceExtractor extends DataSourceJsonCodec`. 
So, same story, parent `DataSourceJsonCodec` are all init, that's an other object so we pay the init cost again, and we reach the point where we reach `NewDataSource` again. 

And here, it tries to init its own `PatchDataSourceHttpType`, but (I think) the init lock for that object is hold by `DataSourceJdbcRepository`, so the field is null. 

Basically, when you use trait, you should be sure that: 
- either you have `val`s but with ONLY ONE direct, linear inheritage line, ie:  ` trait A { ... } ; trait B extends A { ... } ; object C extends B` and that the init order is linear from A to B to C. You can also have several linear hierarchy converging in the last object which "ties the knot". 
- or you use `lazy val` everywhere. Even then, you're not garanted to get all the lock correctly, because scala object lazy init and lock are not proved to work in all case (and they are not, we can trigger the problem easily like here, or here: https://issues.rudder.io/issues/22645 (we has a deadlock, ie the opposite problem), or in scalac bugtracker). So you need to put your object init in a `def`, and it's not very elegant. 

So, I would say for here: 
- don't use trait to hold common data. They are not done to do that, and it leads to surprising init conemdrum, and it duplicate all the data structures (which also means that you will avoid to get jvm nice optim on codec, because depending of the path, it won't be the same object's codec for `DataSourceJsonCodec`.
- if you want, create just one linear hierarchy for Codec, and nothing else. 
- then import the last object which ties the knot where codec are needed, don't extends it.

clarktsiory:
PR updated with a new commit

datasources/src/main/scala/com/normation/plugins/datasources/Serialization.scala
clarktsiory: Thanks for the super clear comment and suggestions ! I like having trait to have that 'hierarchy' and a namespace of codec definitions, but I didn't think about those issues...
I removed the trait as there was no need for it here, I changed it to an `object` :+1:

PR #614, opened 2023-12-04
Fixes #23830: Migrate users and supervised target APIs in change-validation to zio-json

clarktsiory:
https://issues.rudder.io/issues/23830

This is the migration of 2 APIs in the _change-validation_ plugin to use _zio-json_ to replace _lift-json_ (there is one bigger REST API to migrate still, and in consequence lots of refactoring from the lift `Box` type to the zio `IOResult` type in the interfaces, but this will be made in another PR).

I kept the same behavior everywhere as I could, so there may be 'weird' (rather uncommon) things such as logging a decoding error...
I also added tests to the APIs using the _yaml_ format, and to the repositories, which had some saving 'workflow' to test...

### DIFF
--- a/change-validation/src/test/scala/bootstrap/rudder/plugin/MigrateSupervisedGroupsTests.scala
+++ b/change-validation/src/test/scala/bootstrap/rudder/plugin/MigrateSupervisedGroupsTests.scala
@@ -1,0 +1,82 @@
+package bootstrap.rudder.plugin
+
+import better.files.File
+import com.normation.JsonSpecMatcher
+import com.normation.plugins.changevalidation.MockSupervisedTargets
+import com.normation.rudder.rest.RestTestSetUp
+import com.normation.zio.UnsafeRun
+import java.nio.file.Files
+import org.junit.runner.RunWith
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+import org.specs2.specification.AfterAll
+
+@RunWith(classOf[JUnitRunner])
+class MigrateSupervisedGroupsTests extends Specification with JsonSpecMatcher with AfterAll {
+  sequential
+
+  val restTestSetUp = RestTestSetUp.newEnv
+
+  val tmpDir = File(Files.createTempDirectory("rudder-test-"))
+
+  override def afterAll(): Unit = {
+    tmpDir.delete()
+  }
+
+  val supervisedFile   = tmpDir / "supervised-targets.json"
+  val unsupervisedFile = tmpDir / "unsupervised-targets.json"
+
+  val mockNodeGroups = restTestSetUp.mockNodeGroups
+  val mockServices   = new MockSupervisedTargets(
+    tmpDir,
+    "unsupervised-targets.json",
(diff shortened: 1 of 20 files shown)


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

change-validation/src/test/scala/bootstrap/rudder/plugin/MigrateSupervisedGroupsTests.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.normation.zio.UnsafeRun
+import java.nio.file.Files
+import org.junit.runner.RunWith
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+import org.specs2.specification.AfterAll
+
+@RunWith(classOf[JUnitRunner])
+class MigrateSupervisedGroupsTests extends Specification with JsonSpecMatcher with AfterAll {
+  isolated
```
fanf, line 16: it needs to be just `sequential`, not `isolated`, else the whole set up is done again, which is costly and can lead to error depending of the millesecond of directory creation, for ex: 

image

fanf: (previously I got a "directory already exist", which was more explicit)

change-validation/src/test/scala/com/normation/plugins/changevalidation/UnsupervisedTargetsRepositoryTest.scala
```diff
@@ -0,0 +1,70 @@
...
+      file.contentAsString must equalsJsonSemantic("""{"unsupervised":[]}""")
+    }
+
+    val tmpDir     = File(Files.createTempDirectory("rudder-test-"))
+    val tmpDirPath = tmpDir.path
+    val file       = tmpDir / "unsupervised-targets.json"
+
+    "do nothing if the directory is not writable" in {
+      val unsupervisedRepo = new UnsupervisedTargetsRepository(tmpDirPath, "unsupervised-targets.json")
+      tmpDir.setPermissions(PosixFilePermissions.fromString("r--r--r--").asScala.toSet)
```
fanf, line 40: this need to be at least `rw-r--r--` else the test modifying it fails.  (`SupervisedTargetsApiTest`, the `POST`)

clarktsiory: That's right the tests were not passing ! The change was there : https://github.com/Normation/rudder-plugins/blob/31df6b8475ea9c5ae8e6de8e31c93a6b213ef2e0/change-validation/src/test/scala/com/normation/plugins/changevalidation/api/SupervisedTargetsApiTest.scala#L28 (I think I changed it just to see the response I forgot to revert back to `rw` :smile:)

fanf (approved):
LGTM 
(more tests will be needed once rudder header are fully restored)

fanf:
OK, squash merging this PR

PR #616, opened 2023-12-09
Fixes #23858: Error in test (typo)

fanf:
https://issues.rudder.io/issues/23858

Just typo and a little simplification


branding/src/test/scala/com/normation/plugins/branding/api/BrandingApiTest.scala
```diff
@@ -17,8 +17,6 @@ import org.specs2.specification.AfterAll
 class BrandingApiTest extends Specification with TraitTestApiFromYamlFiles with Loggable with AfterAll {
   sequential
 
-  val restTestSetUp = RestTestSetUp.newEnv
```
fanf, line 20: this is rather heavy wait, in that case, I think it's unecessary to pull the whole rest test set up for the quick api test

clarktsiory (approved):
Thanks for the change, LGTM!

PR #617, opened 2023-12-09
Fixes #23857: Impact of API change for tenants on plugins

fanf:
https://issues.rudder.io/issues/23857

Just adapt a few places to match changes in https://github.com/Normation/rudder/pull/5229


fanf:
Auto merging, it's just to correct plugin build

PR #618, opened 2023-12-13
Fixes #23831: Migrate change request API in change-validation to zio-json

clarktsiory:
https://issues.rudder.io/issues/23831

This PR migrates only the zio-json parts of the change request APIs. It consists in : 
* adding tests to the API and the database which is left untouched
* removing usage of `RestExtractorService` from Rudder which is the only part using lift-json indirectly

A whole new ADT tree needed to be created to be in par with the domain model of `ChangeRequest`, and we heavily use _chimney_ to do the mapping, see the [ChangeRequestJson.scala](https://github.com/Normation/rudder-plugins/blob/eebed5083c2ba840ef6919314938d7e5b8178c4b/change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala) file. This time, we use the `PartialTransformer` from _chimney_ because serialization could result in errors anywhere in the transformations. 

The behavior is slightly changed when encountering errors when attempting to serialize a change request : 
* previously we had serialized an "error string" in place of a serialized `Directive` object to encode the error
* now we fail-fast as early as we can before even serializing (e.g. when fetching the context to serialize the `Directive`) and we accumulate any other serialization errors in the error channel, which is provided by _chimney_ and which we adapt to our app's.

There may still be some works to do to for the migration : 
* the lift `Box` could be rewritten to `IOResult` effect type in most places, even up to Rudder interfaces
* the errors messages could be tested and improved further : _chimney_ tracks the path or errors, so when doing the one partial transformation of a `ChangeRequest` we could have very details errors in nested fields.

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala
@@ -1,0 +1,961 @@
+/*
+ *************************************************************************************
+ * Copyright 2023 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package com.normation.plugins.changevalidation
+
+import cats.data.NonEmptyList
+import com.normation.cfclerk.domain.SectionSpec
+import com.normation.cfclerk.domain.Technique
+import com.normation.cfclerk.domain.TechniqueName
+import com.normation.cfclerk.domain.TechniqueVersion
+import com.normation.errors._
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.apidata.JsonResponseObjects._
+import com.normation.rudder.apidata.implicits._
+import com.normation.rudder.domain.nodes.AddNodeGroupDiff
+import com.normation.rudder.domain.nodes.ChangeRequestNodeGroupDiff
+import com.normation.rudder.domain.nodes.DeleteNodeGroupDiff
+import com.normation.rudder.domain.nodes.ModifyNodeGroupDiff
+import com.normation.rudder.domain.nodes.ModifyToNodeGroupDiff
+import com.normation.rudder.domain.nodes.NodeGroup
+import com.normation.rudder.domain.nodes.NodeGroupCategoryId
+import com.normation.rudder.domain.nodes.NodeGroupId
+import com.normation.rudder.domain.policies.AddDirectiveDiff
+import com.normation.rudder.domain.policies.AddRuleDiff
+import com.normation.rudder.domain.policies.ChangeRequestDirectiveDiff
+import com.normation.rudder.domain.policies.ChangeRequestRuleDiff
+import com.normation.rudder.domain.policies.DeleteDirectiveDiff
+import com.normation.rudder.domain.policies.DeleteRuleDiff
+import com.normation.rudder.domain.policies.Directive
+import com.normation.rudder.domain.policies.DirectiveId
+import com.normation.rudder.domain.policies.GroupTarget
+import com.normation.rudder.domain.policies.ModifyDirectiveDiff
+import com.normation.rudder.domain.policies.ModifyRuleDiff
+import com.normation.rudder.domain.policies.ModifyToDirectiveDiff
+import com.normation.rudder.domain.policies.ModifyToRuleDiff
+import com.normation.rudder.domain.policies.Rule
+import com.normation.rudder.domain.policies.RuleId
+import com.normation.rudder.domain.policies.RuleTarget
+import com.normation.rudder.domain.policies.SectionVal
+import com.normation.rudder.domain.policies.SimpleDiff
+import com.normation.rudder.domain.properties.AddGlobalParameterDiff
+import com.normation.rudder.domain.properties.ChangeRequestGlobalParameterDiff
+import com.normation.rudder.domain.properties.DeleteGlobalParameterDiff
+import com.normation.rudder.domain.properties.GenericProperty
+import com.normation.rudder.domain.properties.GlobalParameter
+import com.normation.rudder.domain.properties.GroupProperty
+import com.normation.rudder.domain.properties.InheritMode
+import com.normation.rudder.domain.properties.ModifyGlobalParameterDiff
+import com.normation.rudder.domain.properties.ModifyToGlobalParameterDiff
+import com.normation.rudder.domain.properties.PropertyProvider
+import com.normation.rudder.domain.queries.Query
+import com.normation.rudder.domain.workflows.ChangeRequest
+import com.normation.rudder.domain.workflows.ChangeRequestId
+import com.normation.rudder.domain.workflows.ConfigurationChangeRequest
+import com.normation.rudder.domain.workflows.DirectiveChange
+import com.normation.rudder.domain.workflows.DirectiveChanges
+import com.normation.rudder.domain.workflows.GlobalParameterChange
+import com.normation.rudder.domain.workflows.GlobalParameterChanges
+import com.normation.rudder.domain.workflows.NodeGroupChange
+import com.normation.rudder.domain.workflows.NodeGroupChanges
+import com.normation.rudder.domain.workflows.RuleChange
+import com.normation.rudder.domain.workflows.RuleChanges
+import com.normation.rudder.domain.workflows.WorkflowNodeId
+import com.normation.rudder.services.modification.DiffService
+import com.typesafe.config.ConfigValue
+import io.scalaland.chimney.PartialTransformer
+import io.scalaland.chimney.Transformer
+import io.scalaland.chimney.partial.Result
+import io.scalaland.chimney.syntax._
+import net.liftweb.common.Empty
+import net.liftweb.common.Failure
+import net.liftweb.common.Full
+import scala.util.Try
+import zio.Chunk
+import zio.NonEmptyChunk
+import zio.json._
+import zio.json.internal.Write
+
+sealed trait SimpleDiffOrValueJson[T] {
+  def map[U](f: T => U): SimpleDiffOrValueJson[U]
+}
+final case class SimpleDiffJson[T](
+    @jsonField("from") oldValue: T,
+    @jsonField("to") newValue:   T
+) extends SimpleDiffOrValueJson[T] {
+  def map[U](f: T => U): SimpleDiffJson[U] = SimpleDiffJson(f(oldValue), f(newValue))
+}
+final case class SimpleValueJson[T](value: T) extends SimpleDiffOrValueJson[T] {
+  def map[U](f: T => U): SimpleValueJson[U] = SimpleValueJson(f(value))
+}
+
+object SimpleDiffOrValueJson {
+  implicit def diffTransformer[T]:                 Transformer[SimpleDiff[T], SimpleDiffJson[T]] =
+    Transformer.derive[SimpleDiff[T], SimpleDiffJson[T]]
+  implicit def simpleValueEncoder[T: JsonEncoder]: JsonEncoder[SimpleValueJson[T]]               =
+    JsonEncoder[T].contramap[SimpleValueJson[T]](_.value)
+  implicit def simpleDiffEncoder[T: JsonEncoder]:  JsonEncoder[SimpleDiffJson[T]]                = DeriveJsonEncoder.gen[SimpleDiffJson[T]]
+
+  // An encoder for both {from, to} json object and raw value json. This is a way to encode both different types of json
+  implicit def encoder[T: JsonEncoder]: JsonEncoder[SimpleDiffOrValueJson[T]] = new JsonEncoder[SimpleDiffOrValueJson[T]] {
+    override def unsafeEncode(a: SimpleDiffOrValueJson[T], indent: Option[Int], out: Write): Unit = {
+      a match {
+        case SimpleDiffJson(from, to) => simpleDiffEncoder[T].unsafeEncode(SimpleDiffJson(from, to), indent, out)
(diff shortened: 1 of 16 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

change-validation/src/test/scala/com/normation/plugins/changevalidation/api/ChangeRequestApiTest.scala
```diff
@@ -0,0 +1,659 @@
...
+                      TechniqueName("packageManagement"),
+                      mockDirectives.directives.pkgDirective.copy(
+                        name = "pkg_directive_001",
+                        shortDescription = "testing directive change",
+                        longDescription = "testing directive change",
+                        priority = 1,
+                        isSystem = true,
+                        parameters = mockDirectives.directives.pkgDirective.parameters ++ Map(("PACKAGE_LIST", Seq("curl")))
+                        // ignored changes :
+                        // policyMode = Some(PolicyMode.Audit),
```
fanf, line 299: policy mode should not be ignored. Why is it so ? 
(even tag, I'm not sure) (they should, they are in eventlog: https://issues.rudder.io/issues/15019)

change-validation/src/test/scala/com/normation/plugins/changevalidation/api/ChangeRequestApiTest.scala
```diff
@@ -0,0 +1,659 @@
...
+                      name = "rule_002",
+                      shortDescription = "testing rule change",
+                      longDescription = "testing rule change",
+                      directiveIds = Set(DirectiveId(DirectiveUid("directive1"))),
+                      targets = Set(AllTargetExceptPolicyServers),
+                      isEnabledStatus = false,
+                      isSystem = false
+                      // ignored changes :
+                      // categoryId = RuleCategoryId("rule_category_001"),
+                      // policyMode = Some(PolicyMode.Audit),
```
fanf, line 486: same: I don't see why these change are not tracked. PolicyMode at least has clear impact on prod. 
Can you look to that and potentially open a ticket in the correct branch if needed ?

change-validation/src/main/scala/com/normation/plugins/changevalidation/CurrentUser.scala
```diff
@@ -0,0 +1,90 @@
...
+import com.normation.rudder.AuthorizationType
+import com.normation.rudder.Rights
+import com.normation.rudder.web.services.RudderUserDetail
+import net.liftweb.http.SessionVar
+import org.springframework.security.core.context.SecurityContextHolder
+
+/**
+  * Redefine the Rights methods as in Rudder, constructor allows to provide RudderUserDetail
+  */
+class CurrentUser protected (get: => Option[RudderUserDetail]) extends SessionVar[Option[RudderUserDetail]](get) {
```
fanf, line 49: You absolutly don't want to define these kind of security/authorization thing at several place. If it exists in Rudder, you must not ever define it elsewhere, it's a sure path toward inconsistancy and security hole. 

If you are missing something with Current user: 
- either you change it in Rudder, and ensure that everything still work as expected, 
- or you use adapt for test (if it's only a proble of test). 

Plus: Rudder `CurrentUser` has a `set` method, so you can `CurrentUser.set(Some(userDetails))` if needed

change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala
```diff
@@ -0,0 +1,961 @@
...
+    .withFieldComputed(_.id, _.id.serialize)
+    .withFieldRenamed(_.name, _.displayName)
+    .withFieldComputed(_.categoryId, _.categoryId.value)
+    .withFieldComputed(_.directives, _.directiveIds.map(_.serialize).toList.sorted)
+    .withFieldComputed(_.targets, _.targets.toList.sortBy(_.target).map(t => JRRuleTarget(t)))
+    .withFieldRenamed(_.isEnabledStatus, _.enabled)
+    .withFieldComputed(_.tags, x => JRTags.fromTags(x.tags))
+    .withFieldConst(_.policyMode, None)      // not needed
+    .withFieldConst(_.status, None)          // not needed
+    .withFieldConst(_.changeRequestId, None) // not needed
```
fanf, line 490: I think they should. If they are not in 7.3/8.0, it's a bug - can you check and do what is needed accordingly ?

change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala
```diff
@@ -0,0 +1,961 @@
...
+import io.scalaland.chimney.syntax._
+import net.liftweb.common.Empty
+import net.liftweb.common.Failure
+import net.liftweb.common.Full
+import scala.util.Try
+import zio.Chunk
+import zio.NonEmptyChunk
+import zio.json._
+import zio.json.internal.Write
+
```
fanf, line 112: That seems to have been a really easy piece of code to craft :D

change-validation/src/main/scala/com/normation/plugins/changevalidation/NotificationService.scala
```diff
@@ -113,11 +113,29 @@ class EmailNotificationService {
...
+}
+
+object NotificationService {
+  def apply(
+      emailService:   EmailNotificationService,
+      linkUtil:       LinkUtil,
+      configMailPath: String
+  ): NotificationService = {
+    new NotificationServiceImpl(
+      emailService,
```
fanf, line 127: I think I'm against that idea - but I'm not sure to understand the goal. Is it a kind of default implementation ? 
It looks like it adds a lot of burden in maintenance (each change in constructor have to been done at one more place) and I fail to see the earning. Can you explain ?

change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowJdbcRepository.scala
```diff
@@ -63,36 +63,60 @@ trait WoWorkflowRepository {
...
+  def createWorkflowSQL(crId: ChangeRequestId, state: WorkflowNodeId): Update0 = {
+    sql"insert into workflow (id, state) values ($crId, $state)".update
+  }
+
+  def updateStateSQL(crId: ChangeRequestId, from: WorkflowNodeId, state: WorkflowNodeId): Update0 = {
+    sql"update workflow set state = $state where id = $crId".update
+  }
+}
+trait WorkflowJdbcRepositorySQL extends RoWorkflowJdbcRepositorySQL with WoWorkflowJdbcRepositorySQL {}
+object WorkflowJdbcRepositorySQL extends WorkflowJdbcRepositorySQL
```
fanf, line 90: Can you move `WorkflowJdbcRepositorySQL` into test since it's not use here?
Plus, I think the trait `WorkflowJdbcRepositorySQL` isn't use, you could just have in the test file `object WorkflowJdbcRepositorySQL extends RoWorkflowJdbcRepositorySQL with WoWorkflowJdbcRepositorySQL {}` no ?

change-validation/src/main/scala/com/normation/plugins/changevalidation/api/ChangeRequestApi.scala
```diff
@@ -93,87 +96,94 @@ object ChangeRequestApi       extends ApiModuleProvider[ChangeRequestApi] {
...
     readChangeRequest:    RoChangeRequestRepository,
     writeChangeRequest:   WoChangeRequestRepository,
     readWorkflow:         RoWorkflowRepository,
-    writeWorkflow:        WoWorkflowRepository,
-    readTechnique:        TechniqueRepository,
     workflowLevelService: WorkflowLevelService,
     commitRepository:     CommitAndDeployChangeRequestService,
-    restDataSerializer:   RestDataSerializer
+    userPropertyService:  UserPropertyService,
+    currentUser:          CurrentUser
```
fanf, line 166: you should not pass `CurrentUser` in a constructor. `CurrentUser` is an entity (a strange one, with a very strange persistance and retrieval model) but it's still an entity. 
Only parameters or other services are passed to service constructor. If you want to pass something to get current user (which is ok, it's not great to directly refer a magic entity in code), you should pass `com.normation.rudder.UserService`

change-validation/src/main/scala/com/normation/plugins/changevalidation/api/ChangeRequestApi.scala
```diff
@@ -277,45 +264,14 @@ class ChangeRequestApiImpl(
...
-      } catch {
-        case e: Exception =>
-          toJsonError(None, JString(s"'${sid}' is not a valid change request id (need to be an integer)"))(
-            "changeRequestDetails",
-            restExtractor.extractPrettify(req.params)
-          )
-      }
+      withChangeRequestContext(sid, params, schema, "find")((changeRequest, status, techniqueByDirective) =>
+        serialize(changeRequest, status)(techniqueByDirective).toIO
+      ).toLiftResponseOne(params, schema, Some(sid))
```
fanf, line 269: 👌

change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala
clarktsiory: Here specifically these fields are used when there is a "new" Rule change request or a "deletion" of a Rule change request (`JRRule` is only used for serializing the resulting CR in both cases). 

The previous lift-json code replaced by this one does not in fact serialize these fields : 
```
  def serializeRule(rule: Rule, crId: Option[ChangeRequestId]): JValue = {
    (("changeRequestId"   -> crId.map(_.value.toString))
    ~ ("id"               -> rule.id.uid.value)
    ~ ("revision"         -> rule.id.rev.toOptionString)
    ~ ("displayName"      -> rule.name)
    ~ ("categoryId"       -> rule.categoryId.value)
    ~ ("shortDescription" -> rule.shortDescription)
    ~ ("longDescription"  -> rule.longDescription)
    ~ ("directives"       -> rule.directiveIds.toList.map(_.serialize).sorted)
    ~ ("targets"          -> rule.targets.map(_.toJson))
    ~ ("enabled"          -> rule.isEnabledStatus)
    ~ ("system"           -> rule.isSystem)
    ~ ("tags"             -> serializeTags(rule.tags)))
  }
```

A way to avoid this confusion is the create another case class like `JRRule` but with only those serialized fields, but I chose to reuse `JRRule`, it has all we need (even more)

change-validation/src/test/scala/com/normation/plugins/changevalidation/api/ChangeRequestApiTest.scala
clarktsiory: Both `policyMode` and `tags` changes would be taken into account : https://github.com/Normation/rudder/blob/master/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/DirectiveDiff.scala#L86, but in the API response (yaml tests), the diff for both of them were never displayed : https://github.com/Normation/rudder/blob/master/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/RestDataSerializer.scala#L535

It could be added without risk, but I didn't because it would mean I'd also have to add these new fields to the 'json datatype' and also change the json tests... Is it okay to keep those lines as comments or should I make the change ?

change-validation/src/test/scala/com/normation/plugins/changevalidation/api/ChangeRequestApiTest.scala
clarktsiory: Same answer too, the diff is actually taken into account, here the previous version did not 'serialize' the change for these fields, so I kept that behavior, but it could be changed without risk

change-validation/src/main/scala/com/normation/plugins/changevalidation/NotificationService.scala
clarktsiory: As I wrote a new interface `NotificationService` to be able to use a mock version of it in tests, I renamed the class to `NotificationServiceImpl`, so previously `new NoficationService(args...)` would be now just `NotificationService(args...)`. It was either `NotificationService(args...)` everywhere, or `new NotificationServiceImpl(args...)`. I prefered the first version as it was returning an actual `NotificationService`, the trait and not the implementation.  

It also can be seen as a constructor that hides the implementation, but now the impl remains public (because the actual implementation is tested, including methods not in the interface...) so it kinda defeats its own purpose because of that :sweat_smile:

clarktsiory:
PR updated with a new commit

change-validation/src/main/scala/com/normation/plugins/changevalidation/CurrentUser.scala
clarktsiory: :+1: I didn't think the `set` method existed, it simplifies a lot of things ! (morale : sometimes mutability does, and we need it sometimes) :smile:

Now I only have to wait for this PR to make the `set` method available : https://github.com/Normation/rudder/pull/5267

change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowJdbcRepository.scala
clarktsiory: The solution was as you suggested to remove the unnecessary trait, the `object  WorkflowJdbcRepositorySQL` is used in the file here (to implement the repos) as well as in query type-checking tests, so there is only a single object that contains all the SQL now : https://github.com/Normation/rudder-plugins/pull/618/commits/3fbb60377fcd6ac1c278b9f60319f44797af6e1f

clarktsiory:
PR rebased

change-validation/src/main/scala/com/normation/plugins/changevalidation/CurrentUser.scala
clarktsiory: Now that that one PR is merged I could not do the test without this other one :smile:  : https://github.com/Normation/rudder/pull/5289 (`set`-ting the user is not possible at top-level, we need to do it in each yaml test so it needs a change in the test trait in Rudder)

clarktsiory:
PR updated with a new commit

change-validation/src/main/scala/com/normation/plugins/changevalidation/CurrentUser.scala
clarktsiory: A simpler and better solution : rely on the `UserService`, which allows indeed to have a user with the correct rights in tests : https://github.com/Normation/rudder-plugins/pull/618/commits/9d45951ef7ded84ca44692a0fd84f2c7aa87f8cd

With that last commit I should have addressed all issues for this PR...

clarktsiory:
PR updated with a new commit

change-validation/src/test/scala/com/normation/plugins/changevalidation/api/ChangeRequestApiTest.scala
clarktsiory: I think I would be better off in another PR, because I tried to implement it and the change needs a consequent adaptation of the current code (adapt the `{"from":...,"to":...}` json serialisation because it does not display 'null', and add corresponding lines of code/tests)

change-validation/src/test/scala/com/normation/plugins/changevalidation/api/ChangeRequestApiTest.scala
clarktsiory: Cf. the comment above (I think I would be better off in another PR)

clarktsiory:
PR updated with a new commit

change-validation/src/main/scala/com/normation/plugins/changevalidation/NotificationService.scala
clarktsiory: Resolved in https://github.com/Normation/rudder-plugins/pull/618/commits/e3e6696e56805494aa10d21a8202cf95284744bf !

clarktsiory:
Still rebasing (lasts rebase hopefully :crossed_fingers:)

fanf (approved):
LGTM.
Not exactly a small pr, I hope the remaining lift rest endpoints will be easier to port

PR #624, opened 2024-01-04
Fixes #23984: Change sequence/traverse in plugins too

fanf:
https://issues.rudder.io/issues/23984

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/DataTypes.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/DataTypes.scala
@@ -167,7 +167,7 @@ object Ser {
      */
     for {
       list    <- parseSupervised(json)
-      targets <- Control.sequence(list)(parseTargetId)
+      targets <- Control.traverse(list)(parseTargetId)
     } yield {
       targets.toSet
     }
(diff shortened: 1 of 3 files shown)


clarktsiory (commented):
Shouldn't we also fix warnings ? The plugin does not build (now that we have the `-Xfatal-warnings` option) because of these :

```
[WARNING] /home/clarkenciel/IdeaProjects/rudder-plugins/change-validation/src/main/scala/com/normation/plugins/changevalidation/api/ChangeRequestApi.scala:83: warning: Unused import
[WARNING] import net.liftweb.json.JsonAST.JValue
[WARNING]                                 ^
[WARNING] /home/clarkenciel/IdeaProjects/rudder-plugins/change-validation/src/main/scala/com/normation/plugins/changevalidation/api/ChangeRequestApi.scala:237: warning: local method concatenateJArray in method process0 is never used
[WARNING]           def concatenateJArray(a: JArray, b: JArray): JArray = {
[WARNING]               ^
[WARNING] 2 warnings
```

fanf:
strange, I didn't got it with "make"

fanf:
I don't have it in intellj either o_O

fanf:
Which is normal since it's used in 7.3 - on what branch do you get the warning ?

clarktsiory:
Oh it's because the plugin was not using my locally built Rudder version, but indeed those warnings are not actually there, it's just that they are unused since some code above uses `sequence` which is still `boxSequence` in the Rudder version

fanf:
ok, but there is problems in master, I will see if they are trivial and can be fixed during upmerge, or need dedicated issues

fanf:
So, merging for consistency now that parent is merged

clarktsiory (approved):
LGTM, and it builds successfully !

fanf:
OK, merging this PR

PR #642, opened 2024-02-01
Fixes #24127: Let the plugins Makefile build target without including them in the resulting rpkg

Fdall:
https://issues.rudder.io/issues/24127

Solution: split apport the declaration of files from the declaration of targets


fanf (approved):
LGTM

PR #643, opened 2024-02-02
Fixes #24114: Migrate user-management API to zio-json

clarktsiory:
https://issues.rudder.io/issues/24114


clarktsiory:
PR updated with a new commit

user-management/pom-template.xml
```diff
@@ -48,6 +48,12 @@
 
   <dependencies>
     <!-- Add other plugin specific dependencies -->
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>servlet-api</artifactId>
+      <version>2.5</version>
```
fanf, line 54: don't we use: 
```
      <groupId>javax.servlet</groupId>
      <artifactId>javax.servlet-api</artifactId>
      <version>3.1.0</version>
      <scope>provided</scope>
```
In other projects, we should/need to homogeneize that. Perhaps we should just put it in the parent pom since it seems we need it everywhere.

clarktsiory:
PR updated with a new commit

user-management/pom-template.xml
clarktsiory: Totally ! We used 3.1.0 as well when migrating other plugins
I don't know where I copied that from...

:white_check_mark: https://github.com/Normation/rudder-plugins/pull/643/commits/63834f6ac97500bd786b8532175a646fae902398

fanf (approved):
Everything works, new tests are well, the resulting code is cleaner and faster, GG!

clarktsiory:
PR updated with a new commit

PR #651, opened 2024-02-09
Fixes #24182: Add a role mapping and filtering for OIDC provided roles

fanf:
https://issues.rudder.io/issues/24182

This PR provides two main things: 
- a way to map IdP provided roles to rudder internal roles, and optionnally to restrict IdP addressable roles to only the mapped one
- configuration option to use IdP Single Log Out by configuring a logout url (and post logout redirect url)

The code is not very interesting and mostly deals with adding option, reading them, logs and spring workaround. 

For role, we just have a new layer of mapping that uses the `Alias` role. 
For redirect, we just have a new `PostLogoutAction` that build the correct URL (the main changes are in rudder).


clarktsiory (commented):
LGTM ! There is an error at startup on my side for the bean `UserRepository` which cannot be found (when instantiating the `oidcUserService` bean), to make it work I just added : 
```scala
@Bean def userRepository: UserRepository = RudderConfig.userRepository
```

PR #652, opened 2024-02-09
Fixes #24184: User management roles and permissions are intermingled in the UI

clarktsiory:
https://issues.rudder.io/issues/24184

Basically we change the API format to separate _roles_ and _authorizations_, and also to add some fields, now the meaning of the fields are :
* `permissions` : roles only, without any authorizations
* `authz` : all authorizations (or "rights" if you prefer) associated with the `permissions`
* `rolesCoverage` : all inferred roles from all user authorizations, a superset of `permissions`
* `customRights` : set of custom authorizations that were manually added (not related to any `permissions`)

In the UI we add a "readonly section" (below the "Delete" and "Save" buttons to edit the user), in which we display the authorizations and the inferred roles (the roles can still be edited in the same field as before) :

image

Now the UI is more consistent, we can remove only remove roles from the UI, and the related information will update accordingly.
But we still need to add/remove custom rights from the users file, it's the desired behavior because ideally they should be defined in custom roles.

### DIFF
--- a/user-management/src/main/elm/sources/ApiCalls.elm
+++ b/user-management/src/main/elm/sources/ApiCalls.elm
@@ -6,10 +6,10 @@ module ApiCalls exposing (..)
 -- API call to get the category tree
 
 
-import DataTypes exposing (AddUserForm, Authorization, Model, Msg(..), User)
+import DataTypes exposing (AddUserForm, Model, Msg(..))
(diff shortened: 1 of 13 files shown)


clarktsiory:
PR updated with a new commit

fanf:
So much clearer :+1:

PR #655, opened 2024-02-14
Fixes #24189: No API right with aliased roles

clarktsiory:
https://issues.rudder.io/issues/24189

We copied only the `roles` to override the user details from `Rudder` : https://github.com/Normation/rudder/blob/backports/7.3.12/24146/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala#L386.

But we also need to replace the `ApiAuthorization.None` by authorizations computed from the set of roles


fanf (approved):
:ok_hand:

PR #656, opened 2024-02-14
Fixes #24191: Make version of backport branch ok for rudder-package

fanf:
https://issues.rudder.io/issues/24191

Add a variable for the version to use in rpdk metadata so that rudder package is happy. It's not a solid way to do it, we should avoid to port it to 7.3 branch when merging - but at least, it will be easy to find/correct if merged in main branches by error.


clarktsiory (approved):
:ok_hand:

PR #658, opened 2024-02-16
Fixes #24202: No API right with OIDC provided roles

clarktsiory:
https://issues.rudder.io/issues/24202

This is a port of https://github.com/Normation/rudder-plugins/pull/655 to 7.3 branch


auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
```diff
@@ -285,6 +288,8 @@ class AuthBackendsSpringConfiguration extends ApplicationContextAware {
     }
   }
 
+  @Bean def roleApiMapping = RudderConfig.roleApiMapping
+
```
fanf, line 292: This need to be changed to 'new RoleApiMapping' b/c if not, it needs a new version of rudder-core.
we will do that full correction in the backport (and then merge it)

clarktsiory:
PR updated with a new commit

VinceMacBuche (approved):

PR #659, opened 2024-02-16
Fixes #24206: We need an option to force validation of change requests

clarktsiory:
https://issues.rudder.io/issues/24206

We use the configuration from https://github.com/Normation/rudder/pull/5404.

The checkbox behaves the same as the ones for enabling/disabling change requests :
image

### DIFF
--- a/change-validation/src/main/scala/bootstrap/rudder/plugin/ChangeValidationConf.scala
+++ b/change-validation/src/main/scala/bootstrap/rudder/plugin/ChangeValidationConf.scala
@@ -44,7 +44,6 @@ import bootstrap.liftweb.RudderConfig.restDataSerializer
 import bootstrap.liftweb.RudderConfig.restExtractorService
 import bootstrap.liftweb.RudderConfig.techniqueRepository
 import bootstrap.liftweb.RudderConfig.workflowLevelService
-
 import com.normation.box._
 import com.normation.eventlog.EventActor
 import com.normation.plugins.PluginStatus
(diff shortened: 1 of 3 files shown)


change-validation/src/main/scala/bootstrap/rudder/plugin/ChangeValidationConf.scala
```diff
@@ -144,7 +143,7 @@ class ChangeValidationWorkflowLevelService(
       change:  T
   ): Box[WorkflowService] = {
     def getWorkflowAux = {
-      getWorkflow(validationNeeded.foldLeft(Full(false): Box[Boolean]) {
+      getWorkflow(validationNeeded.foldLeft(alwaysNeedValidation()) {
```
clarktsiory, line 146: this is the main change

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

VinceMacBuche:
I would have preferred the setting within plugin directly. other settings were very old but i think we should not do likewise. Was that decided with @fanf ?

clarktsiory:
PR updated with a new commit

fanf:
OK, squash merging this PR

VinceMacBuche:
I really wanted to not have new change related settings in Rudder but nevermind

VinceMacBuche:
my question was not about the UI, but the backend storage of the value that is in Rudder.

PR #667, opened 2024-02-26
Fixes #24247: Authentication protocol names should be normalized to lowercase

clarktsiory:
https://issues.rudder.io/issues/24247

### DIFF
--- a/auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
+++ b/auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
@@ -265,11 +265,11 @@ class AuthBackendsSpringConfiguration extends ApplicationContextAware {
       // add authentication providers to rudder list
 
       RudderConfig.authenticationProviders.addSpringAuthenticationProvider(
-        "oauth2",
+        RudderOAuth2UserService.PROTOCOL_ID,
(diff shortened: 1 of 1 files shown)


clarktsiory:
PR updated with a new commit

auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
```diff
@@ -616,7 +616,7 @@ class RudderOidcUserService(
   // we need to use our copy of DefaultOAuth2UserService to log/manage errors
   super.setOauth2UserService(new RudderDefaultOAuth2UserService())
 
-  override val protocolName = "OIDC"
+  override val protocolName = "oidc"
```
fanf, line 619: This is a human readable name used in debug/log, so this one should be as it is in the litterature. 
The problem is likely that we should not use that for the thing we put in the database.

fanf: I think what we want is the key of the authentication provider stored in `RudderConfig.authenticationProviders` used for the authentication

fanf (approved):
Neat!

fanf (approved):

fanf:
Merging by hand because feature branch

PR #668, opened 2024-02-26
Fixes #24219: Display user detail using information from database

clarktsiory:
https://issues.rudder.io/issues/24219

The main UI change in this PR consists in handling users who have multiple provider information sources, and setting the correct UI state accordingly : 
* a user can have a provider which "extends" the `file` provider, so we need to display both roles inherited from the provider (readonly) and from the file (can be changed) : 
image

* a user can have a provider which "overrides" roles so ignores the file completely, so the roles cannot be changed at all and only inherited from the provider : 
image
  
Since we can get query provider information in https://github.com/Normation/rudder/pull/5408, we can pass them as json in the API such as : 

```
{
  "providerProperties": {
    "ldap": {
      "roleListOverride": "none",
      "hasModifiablePassword": true
    },
    "oidc": {
      "roleListOverride": "extend",
      "hasModifiablePassword": false
    }
  }
}
```
for the UI to handle the logic depending on the value of `roleListOverride` for the provider.

We also need to have the user roles defined by each provider (see the yml files for complete examples) : 
```
{
   ...,
   "users": [
     {
        "id": "test user",
        "roles": ["inventory", "read_only"],
        "providersInfo": {
          "oidc": {
            "roles": ["inventory"],
            ...
           },
           "file": {
             "roles": ["read_only"],
             ...
           }
         }
       }
   ]
}
```


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

user-management/src/main/scala/com/normation/plugins/usermanagement/UserManagementService.scala
```diff
@@ -222,14 +251,15 @@ class UserManagementService(userRepository: UserRepository, getUserResourceFile:
...
                         Unexpected(s"Wrong formatting : ${file.path}").fail
                     }
+      _          <- userService.reloadPure()
     } yield user
   }
 
   /*
    * When we delete an user, it can be from file or auto-added by OIDC or other backend supporting that.
-   * So we need to both remove it (mark "status=delete") from base and from file.
+   * So we let the callback on file reload do the deletion for file users, and then delete from the user repository.
```
clarktsiory, line 260: This is a tricky bug that relates to how the "reload" work, the callback behind the deletion does some logic on users managed by file that can also change other users.
We absolutely don't want this logic for users managed by other providers, and we wait for the callbacks to terminate before effectively setting the status to deleted.

clarktsiory:
PR rebased

PR #673, opened 2024-02-28
Fixes #24281: Modify and show user information from database

clarktsiory:
https://issues.rudder.io/issues/24281

:warning: **THIS IS STILL ON TOP OF https://github.com/Normation/rudder-plugins/pull/668**

Allowing to change `name`, `email` and to add additional fields which will be stored as key value strings in `otherinfo`.

This needed some refactoring on the user form state within the model in Elm. Validation of those fields are not implemented, and could be done later, for now the information is only used for display. 

Also the last login is displayed at the bottom of the user card and at the head of the user edit panel : 
image


user-management/src/main/scala/com/normation/plugins/usermanagement/api/UserManagementApi.scala
```diff
@@ -300,23 +359,110 @@ class UserManagementApiImpl(
...
+                                  .map {
+                                    case true  =>
+                                      implicit val currentRoles   = allRoles.values.toSet
+                                      // Unknown permissions are trusted and put in file
+                                      val (roles, authz, unknown) = UserManagementService.parsePermissions(user.permissions.toSet)
+                                      userManagementService
+                                        .add(User(u.id, "", roles.map(_.name) ++ authz.map(_.id) ++ unknown), user.isPreHashed)
+                                    case false =>
+                                      Inconsistency(
+                                        s"Roles of user '$id' cannot be extended by provider '${u.managedBy}' because how the provider was configured"
```
fanf, line 382: This fail if I use `providers = oidc, ldap, file` and try to change the name of an auto provisionned user (ie one without existence in `rudder-user.xml`) but works with an user that exists.  
I think there is two things to change here: 
- 1/ I think the role extension is misunderstood and partially mixed with autoprovisioning. 

An user can always be declared in the `rudder-user.xml`, even if the OIDC provider is configured with autoprovisioning. It's just that it *doesn't* need to, and if it is not, then only properties from the oidc token will be considered. 

Roles can be not at all (`roles.enabled=false`), partially (`roles.enabled=true`, `roles.override=false`), or totally (``roles.enabled=true`, `roles.override=true`) provided by OIDC. But even in the last case, if several providers are configured for Rudder, then you should be able to set role from the UI b/c the idea is that we have a list of fallback, and so, if Bob tried to log-in with OIDC and that failed (server not available for ex), then he can try with file but in that case, he will get the rights that are defined in `rudder-user.xml`for him. These right can be minimal, for ex read-only, because it's a fall back with lesser level of identification. 
This is also used so that there is an administrator that can still log-in even with a main authentication provider down.

But for now, if it's too complicated, we can accept that an auto-provisioned user (ie one that never reached `rudder-user.xml`), we error on role modification (but it would be better to just, in that case, create the user in the xml file, and warn the admin that if oidc is configured for override, what it does will only matter for the fallback case). 

- 2/ I think we need to have two separated form in the user edit: 
    - one for the things in data base (name, email, other fields)
      This one doesn't have anything to do with the authorisation config and is simple crud on base fields
    - one for roles
      The one with these logic checks



Plus, the error is not correctly handled in the elm app, we just get a white page with a red "Error" toast notification

user-management/src/main/scala/com/normation/plugins/usermanagement/api/UserManagementApi.scala
```diff
@@ -185,24 +217,53 @@ class UserManagementApiImpl(
...
+                      // depending on provider property configuration, we should merge or override roles, by default merge
+                      val providerProperties = authProvider.getProviderProperties()
+                      val overrideRoles      = providerProperties.get(u.managedBy).map(_.overridesRoles).getOrElse(false)
+                      // merge at the level of JsonUser because last session roles are just String
+                      val providedUser       = transformProvidedUser(u, lastSession)
+                      if (overrideRoles) {
+                        providedUser.addProviderInfo(fileProviderInfo)
+                      } else {
+                        providedUser.merge(fileProviderInfo).withRoleCoverage(x)
+                      }
```
fanf, line 258: There is something strange with that API endpoint: I try to log/unlog/relog, and still get "never log in"
Plus I thing we want to adapt text to "first log-in", else it's a bit confusing

image

fanf: AH ! Perhaps I have an idea:
- the first log-in doesn't have an previous sessions
- there seems to be a bug when the user is auto logged-out because sessions expired, the base is not correctly filled - so still no closed sessions
- and finally when I logged out before sessions expired and logged in again, I got the correct behavior 

image

clarktsiory: It's because the query for the previous login adds a `where enddate is not null`, maybe we should parameterize that predicate ? 
It's useful [here](https://github.com/Normation/rudder/blob/backports/7.3.12/24146/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/UserInformation.scala#L77) where we want the current user to know the last time they logged in.
But in user-management we want the 'current' session to be considered

PR #675, opened 2024-02-29
Fixes #24296: Log on user api authorizations should use a debug string for ACL

clarktsiory:
https://issues.rudder.io/issues/24296

This depends on the `debugString` method introduced in https://github.com/Normation/rudder/pull/5414


fanf (approved):
Lgtm, will merge by hand, feature branch

PR #692, opened 2024-03-18
Fixes #24521: Remove ansible-policies plugin

amousset:
Hypothetical current users having the plugin installed will just keep the technique, so it won't break anything. It was also apparently never released but only provided as nightly.

https://issues.rudder.io/issues/24521


fanf (approved):
OK, let's get ride of it

PR #694, opened 2024-03-19
Fixes #24533: Update to scala 3 syntax - branch 7.3 - public plugins

fanf:
https://issues.rudder.io/issues/24533


fanf:
Merging by hand so that I can handle the upmerge

PR #699, opened 2024-03-29
Fixes #24623: Several compilation error following change in rudder-core

fanf:
https://issues.rudder.io/issues/24623

Nothing really interesting, just removing removed methods from trait and moving a group in a subcategory


fanf:
PR updated with a new commit

VinceMacBuche (approved):

PR #706, opened 2024-04-23
Fixes #24771: Update scala plugin dependencies

clarktsiory:
https://issues.rudder.io/issues/24771

Minor bumps for some plugins : 
* `auth-backends` : :man_dancing: jackson-databind 
* `change-validation` : :incoming_envelope: new version uses the jakarta mail 2 API now 
* `openscap` : :wastebasket: neko-html-unit can be deleted as stated in the comment

### DIFF
--- a/api-authorizations/pom-template.xml
+++ b/api-authorizations/pom-template.xml
@@ -49,7 +49,7 @@
     <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>servlet-api</artifactId>
-      <version>2.5</version>
+      <version>3.1.0</version>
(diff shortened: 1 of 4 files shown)


api-authorizations/pom-template.xml
```diff
@@ -49,7 +49,7 @@
     <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>servlet-api</artifactId>
-      <version>2.5</version>
```
clarktsiory, line 52: This should be 3.1.0, the actual version provided by Rudder

fanf (approved):

PR #715, opened 2024-05-14
Fixes #24781: OpenSCAP technique doesn't show up in GUI after OpenSCAP plugin installation

ncharles:
https://issues.rudder.io/issues/24781


ncharles:
PR updated with a new commit

Fdall (approved):

PR #723, opened 2024-06-06
Fixes #24979: API accounts cannot be changed when api-authorizations plugin is enabled

clarktsiory:
https://issues.rudder.io/issues/24979

The `Accounts` Elm app in Rudder creates a new modal component `modal-account` every time we edit an account, and the modal get deleted from the DOM every time a new account is rendered on the page.
Within the modal DOM element, there is a `<div id="apiauthorization-content"/>` which is a the target of the mounting of the `ApiAuthorizations` Elm app that displays the list of ACLs. 

The problem is that the `ApiAuthorizations` Elm app is no longer mounted, because the modal system has changed and mounting is attempted before the target is available.  

To fix it in the plugin we need to :
* await the modal creation, to guarantee the presence of `<div id="apiauthorization-content"/>`
* do this every time we display an account : we need an Elm app for each API account now

### DIFF
--- a/api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/ApiAccountsExtension.scala
+++ b/api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/ApiAccountsExtension.scala
@@ -114,14 +114,33 @@ class ApiAccountsExtension(val status: PluginStatus)(implicit val ttag: ClassTag
         </head_merge>
         <script>
         //<![CDATA[
+          function waitForElm(selector) {
+            return new Promise(resolve => {
+              if (document.querySelector(selector)) {
+                return resolve(document.querySelector(selector));
+              }
+
+              const observer = new MutationObserver(mutations => {
+                if (document.querySelector(selector)) {
+                  observer.disconnect();
+                  resolve(document.querySelector(selector));
+                }
+              });
+
+              observer.observe(document.body, {
+                childList: true,
+                subtree: true
+              });
+            });
+          }
           var appAcl;
-          $(document).ready(function(){
-            app.ports.initAcl.subscribe(function(){
+          $(document).ready(async function(){
+            async function initAcl(){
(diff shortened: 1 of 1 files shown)


api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/ApiAccountsExtension.scala
```diff
@@ -131,10 +152,18 @@ class ApiAccountsExtension(val status: PluginStatus)(implicit val ttag: ClassTag
...
             //get the acl list of the selected account
-            app.ports.shareAcl.subscribe(function(acl){
-              appAcl.ports.getToken.send(acl)
+            app.ports.shareAcl.subscribe(async function(token){
+              // only initialize with a new token when modal is opened with a known token id :
+              // this is an implicit contract between app and appAcl
+              if (appAcl !== undefined && token.id !== "") {
+                await initAcl(token);
+                appAcl.ports.getToken.send(token)
+              }
```
clarktsiory, line 166: the check for `token.id !== ""` was necessary, because it seems that the `Accounts` app on Rudder make the plugin load an empty token every time (so when an existing one with custom ACLs, it appears with empty acl)

api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/ApiAccountsExtension.scala
```diff
@@ -114,14 +114,35 @@ class ApiAccountsExtension(val status: PluginStatus)(implicit val ttag: ClassTag
         </head_merge>
         <script>
         //<![CDATA[
+          function waitForElm(selector) {
```
clarktsiory, line 117: https://stackoverflow.com/questions/5525071/how-to-wait-until-an-element-exists

VinceMacBuche (approved):

PR #724, opened 2024-06-12
Fixes #24996: UpdateHttpDatasetTest datasource test get blocked forever

fanf:
https://issues.rudder.io/issues/24996

So, there was two problems: 

- 1/ `try` in `attempt` leads to stalling tests

We had that: 

```
        IOResult.attemptZIO(try {
          json.read[JSONAware](path).succeed
        } catch {
          case _: ClassCastException =>
            try { ...
```

And in case of exception, it was not doing what was expected but failing with the exception. It might be due to a change in the way `ZIO.attempt` works (since we switched to 2.x around here), but how the hell didn't we caught it before ? Or is it a more recent change in 2.X branch ? (but then, the last upgrade is in 08/2023 on 8.0 branch).

* 2/ Second problem: we had a `.toString` in the fallback case, so all values, even when they were map (ie json objects), were changed into string, so quote espacted, then parsed back AS STRING. Which obviously does not work. 

And that leads to two questions: 
- if at some point, `JSONAware` stopped to be able to cast `LinkedArrayList` to `JSONObject`, when did it happened (since we don't seems to have change json path version in all 8.0 live)
- if it never worked, then how the `.toString` was added ? It seems to have always been there - since 2019 at least. 

So that's more question than solution. IT REALLY STRANGE and it will be good to find the root change that lead to that.

### DIFF
--- a/datasources/src/main/scala/com/normation/plugins/datasources/UpdateHttpDataset.scala
+++ b/datasources/src/main/scala/com/normation/plugins/datasources/UpdateHttpDataset.scala
@@ -54,10 +54,8 @@ import com.normation.zio.*
 import com.softwaremill.quicklens.*
 import com.typesafe.config.ConfigValue
 import net.minidev.json.JSONArray
-import net.minidev.json.JSONAware
 import net.minidev.json.JSONStyle
 import net.minidev.json.JSONValue
-import scala.util.control.NonFatal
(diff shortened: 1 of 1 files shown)


clarktsiory (approved):
LGTM, and I agree that this is really weird... At least with this cleaner way, it works !

PR #727, opened 2024-06-15
Fixes #25015: Plugins need to be ported, too

fanf:
https://issues.rudder.io/issues/25015

Lots of changes going on here : 
- for all plugins, changing the use of `TraitTestApiFromYamlFiles` ; similar, purely mechanic changes appart in datasources because the test was mixing both that and standard specs2 tests, I splitted it in two files (`RestDataSourceTest.scala` for specs2 and `RestDataSourceFilesTest.scala` for rest YAML)
- some YAML test file have some changes in json property orders. This is due to the fact that the normalisation with the new `TraitTestApiFromYamlFiles` is not exactly the same than before
- some `.apply` are added to comply to more recent scala rules
- `ScaleOutRelay` is adapted to `policyTag`: it seems that it's limited to only removing parameters in  `getBundleVariables` 
- because of changes in node API and policyType, it was simpler to finish migrating datasource to `CoreNodeFact` than to try to add more compatibility layer with `NodeInfo`

Datasources changes: 

- change everywhere `NodeInfoService` and `WoNodeRepository` to `NodeFactRepository`. It tends to simplify things since now we have only one repo for both read and write parts
- change `NodeInfo` to `CoreNodeFact` with the usual impact on some attributes path (`policyServerId`, `hostname`/`fqdn`, etc)
- adapt the check that look if a node is updated to also use what `CoreNodeFactRespository` say with its change event
- then, big big changes in [UpdateHttpDatasetTest.scala](https://github.com/Normation/rudder-plugins/pull/727/files#diff-44d43b7b8b22fabae4537dae04dcf91ea30fe5594278681a49ffe9e43951e4e8)

Datasources > UpdateHttpDatasetTest changes : 
The changes are due to the fact that we use to have a special `NodeInfoService` that is used to track how many time a node property is changed in a side `Map[NodeId, Int]` and avoid having to deal with LDAP. This needed to be reworked because of the change to `CoreNodeFactRepository`, but it's *much simpler* now: 
- we can avoid backend storage mess by configuring a `CoreNodeFactRepository` with `NoopFactStorage` , 
- we can finely track node changes with a `NodeFactChangeEventCallback`, they are for that.

So now, we just update the `updates: Map[NodeId, Int]` in a callback. I also put it into a `Ref`, hoping it will limit the number of instability on that test. 
The remaning changes are implication of that.

### DIFF
--- a/api-authorizations/src/test/scala/com/normation/plugins/apiauthorizations/api/UserApiTest.scala
+++ b/api-authorizations/src/test/scala/com/normation/plugins/apiauthorizations/api/UserApiTest.scala
@@ -1,6 +1,8 @@
 package com.normation.plugins.apiauthorizations
 
 import better.files.*
+import com.normation.errors.IOResult
+import com.normation.errors.effectUioUnit
 import com.normation.rudder.AuthorizationType
 import com.normation.rudder.api.ApiAccount
 import com.normation.rudder.api.ApiAccountId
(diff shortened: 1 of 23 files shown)


fanf:
PR updated with a new commit

fanf:
PR rebased

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR rebased

fanf:
OK, merging this PR

PR #730, opened 2024-06-17
Fixes #25023: Fix update dependencies for datasources and change-validation

clarktsiory:
https://issues.rudder.io/issues/25023
* `change-validation` : The current task for cleaning CSS files in the _gulpfile_ removes all resources files in the directory (precious HTML and JS code), but it should only remove `.css` ones
* `datasources` : The _package.json_ file is not symlinked to `plugins-common`'s file, contrary to other plugins


fanf (approved):
Works as expected !

PR #736, opened 2024-06-28
Fixes #25065: 

clarktsiory:
https://issues.rudder.io/issues/25065

There is nothing to be escaped at all, really...

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/snippet/ChangeRequestChangesForm.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/snippet/ChangeRequestChangesForm.scala
@@ -70,6 +70,7 @@ import net.liftweb.http.*
 import net.liftweb.http.js.JE.*
 import net.liftweb.http.js.JsCmds.*
 import net.liftweb.util.Helpers.*
+import org.apache.commons.text.StringEscapeUtils
 import org.joda.time.DateTime
 import scala.xml.*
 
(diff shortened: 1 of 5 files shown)


change-validation/src/main/scala/com/normation/plugins/changevalidation/snippet/ChangeRequestChangesForm.scala
```diff
@@ -705,7 +706,8 @@ class ChangeRequestChangesForm(
     "#description" #> displaySimpleDiff(diff.modDescription, "description", Text(param.description)))(globalParameterXML)
   }
 
-  private[this] def displayFormDiff[T](diff: SimpleDiff[T], name: String)(implicit fun: T => String = (t: T) => t.toString) = {
+  private[this] def displayFormDiff[T](diff: SimpleDiff[T], rawName: String)(implicit fun: T => String = (t: T) => t.toString) = {
+    val name = StringEscapeUtils.escapeEcmaScript(rawName)
```
clarktsiory, line 710: `name` is supposed to be const everywhere the method is called (e.g. with Scala 3 union types and literal types it would have been obvious from the type that we would'nt need to escape...). I escaped it just in case here

fanf (approved):
A simple PR :sweat_smile:

clarktsiory:
OK, merging this PR

clarktsiory:
OK, merging this PR

PR #737, opened 2024-06-28
Fixes #25085: Datasources tests fail on 404

clarktsiory:
https://issues.rudder.io/issues/25085

noidea

Tests are still unstable, I was not sure of how to reproduce test failures : once I did the change and it passed once, I double-checked the behavior by reverting to the old version but then the tests always passed :weary: 

Before that, I changed the `nThreads` value to 1, and the test passed (but other tests on parallelism failed).
And yet I didn't find any "shared" state within the failing tests on 404 error, so I assumed this is a "client" behavior, and therefore the client IO may not finish before specs2 checks the second branche of the `... and ...`

### DIFF
--- a/datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
+++ b/datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
@@ -1224,13 +1224,15 @@ class UpdateHttpDatasetTest extends Specification with BoxSpecMatcher with Logga
         nodeIds.map(n => if (expectMod) NodeUpdateResult.Updated(n) else NodeUpdateResult.Unchanged(n): NodeUpdateResult)
       )
 
-      res.either.runNow must beRight(matcher) and (
-        if (expectMod) {
-          updates.get.runNow must havePairs(nodeIds.map(x => (x, 1)).toSeq*)
-        } else {
-          updates.get.runNow must beEmpty
-        }
-      ) and ({
+      res.either
+        .flatMap(r => {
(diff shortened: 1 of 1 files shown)


datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala
```diff
@@ -1224,13 +1224,15 @@ class UpdateHttpDatasetTest extends Specification with BoxSpecMatcher with Logga
         nodeIds.map(n => if (expectMod) NodeUpdateResult.Updated(n) else NodeUpdateResult.Unchanged(n): NodeUpdateResult)
       )
 
-      res.either.runNow must beRight(matcher) and (
-        if (expectMod) {
-          updates.get.runNow must havePairs(nodeIds.map(x => (x, 1)).toSeq*)
-        } else {
-          updates.get.runNow must beEmpty
```
clarktsiory, line 1231: In previous branches (8.1 and before), this line is `true must _== true`... I wonder if we shouldn't worry about this and force it like that :sweat_smile:

fanf:
Ex of pb: https://ci.normation.com/jenkins/blue/organizations/jenkins/rudder-plugins/detail/master/1221/pipeline/235

fanf:
Well, removing `runNow` is always a good move, so I'm going to merge that. At worst, it will help in the path toward a single entry point for exec.

fanf (approved):
LGTM
(on my part, I wasn't able to trigger the error neither before nor after :shrug:  )

PR #740, opened 2024-07-25
Fixes #25193: OpenSCAP test fails when there is a '@' '%40' in path

fanf:
https://issues.rudder.io/issues/25193


fanf:
PR updated with a new commit

PR #741, opened 2024-07-25
Fixes #25075: OpenSCAP plugin output displays on *all* node tabs

RaphaelGauthier:
https://issues.rudder.io/issues/25075


fanf (approved):
html is *a bit* fragile, isn't it?

PR #743, opened 2024-08-07
Fixes #25280: Clarify that we do not support authorizations in OIDC authentification role in documentation

ElaadF:
https://issues.rudder.io/issues/25280


ElaadF:
PR rebased

PR #744, opened 2024-08-16
Fixes #25313: Reload users file when user is disabled or activated in plugin

clarktsiory:
https://issues.rudder.io/issues/25313

This is both needed to make the plugin compile, and for changes in https://github.com/Normation/rudder/pull/5816


clarktsiory:
OK, merging this PR

clarktsiory:
OK, merging this PR

PR #775, opened 2024-12-03
Fixes #25999: Disabling user rest API token feature from api-authorizations plugin UI 

clarktsiory:
https://issues.rudder.io/issues/25999

UI for the feature switch in https://github.com/Normation/rudder/pull/6050 : essentially we don't display the user API token menu in the api-authorizations plugin snippet.
We still have an API endpoint to pass the configuration value, also there is some refactoring and formatting applied to the Elm code which knows about the status by calling the endpoint.

In the related PR, the user management roles and default providers are also changed, so there is some changes in the user-management plugin


clarktsiory:
PR updated with a new commit

fanf:
After further consideration, we wan to totally disable (not show) the pop-up when the feature is disabled (globally or for an user), as if the feature was not here.

fanf (changes requested):
After further consideration, we wan to totally disable (not show) the pop-up when the feature is disabled (globally or for an user), as if the feature was not here.

clarktsiory:
PR rebased

fanf (approved):
Perfect !

clarktsiory:
OK, merging this PR

clarktsiory:
OK, merging this PR

PR #779, opened 2024-12-13
Fixes #26065: OpenSCAP report needs to be rendered without iframe with sanitized html

clarktsiory:
https://issues.rudder.io/issues/26065

We have to read the report file (as the API does) in the snippet, as pass the information down, to create buttons with a download action (similar to https://github.com/Normation/rudder/pull/5941).

The iframe can be used to render the original report by putting the sanitized content as string in [the srcdoc attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#srcdoc)

Result : 
Screenshot from 2024-12-13 16-16-40

### DIFF
--- a/openscap/src/main/scala/com/normation/plugins/openscappolicies/extension/OpenScapNodeDetailsExtension.scala
+++ b/openscap/src/main/scala/com/normation/plugins/openscappolicies/extension/OpenScapNodeDetailsExtension.scala
@@ -37,19 +37,25 @@
 
 package com.normation.plugins.openscappolicies.extension
 
-import com.normation.inventory.domain.NodeId
 import com.normation.plugins.PluginExtensionPoint
 import com.normation.plugins.PluginStatus
+import com.normation.plugins.openscappolicies.OpenScapReport
 import com.normation.plugins.openscappolicies.services.OpenScapReportReader
+import com.normation.plugins.openscappolicies.services.ReportSanitizer
 import com.normation.rudder.users.CurrentUser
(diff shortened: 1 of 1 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

openscap/src/main/scala/com/normation/plugins/openscappolicies/extension/OpenScapNodeDetailsExtension.scala
```diff
@@ -129,13 +145,46 @@ class OpenScapNodeDetailsExtension(
...
-    ".original [download]" #> s"OpenSCAP report for ${hostname} (${nodeId.value}).html"
+  private def downloadReportButton(fileName: String, bytes: Array[Byte], sanitized: Boolean) = {
+    val base64   = Base64.getEncoder().encodeToString(bytes)
+    val download = Run(
+      s"""saveByteArray("${fileName}", "application/zip", base64ToArrayBuffer("${base64}"));"""
+    )
+
+    val btn = if (sanitized) {
+      SHtml.ajaxButton(
+        <span><i class="fa fa-download text-light"></i>Download report</span>,
```
fanf, line 175: We don't want a download here, we want to open a tab with the report. That allows to see it in full page, with link navigation available, which is mandatory to be able to see anything.

fanf (changes requested):

clarktsiory:
PR updated with a new commit

fanf (approved):

PR #784, opened 2025-01-13
Fixes #26172: in datasources, ids are not normalized, thus creating problems at deletion

fanf:
https://issues.rudder.io/issues/26172

A LOT of problems: 

- error were not displayed in the main form for empty name and id
- empty URL was not checked for an error
- the ID was not normalized on save like it was displayed in the placehold
- backend says that the duration for schedule is in seconds, but front end uses minutes
- an error was silently absorbed on repository save failure
- repository is not enforcing the same rule than front end on ID allowed chars

Plus cleaning unused things.

### DIFF
--- a/datasources/src/main/elm/sources/DataSourceForm.elm
+++ b/datasources/src/main/elm/sources/DataSourceForm.elm
@@ -15,10 +15,52 @@ datasourceForm : Model -> DataSource -> Maybe DataSource -> Html Msg
 datasourceForm model datasource origin =
   let
 
+    isNew = case origin of
+              Nothing -> True
+              Just o -> False
+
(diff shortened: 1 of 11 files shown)


fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM !
It is easier to use but there are still old UX behavior that we will we to remove later : 
* the Save button is never disabled, even when some fields still need to be filled
* adding a datasource with validation error on the ID (e.g. with spaces) will add the datasource in the list, but show an error notification and refuse datasource creation (it is a frontend problem, which is solved by reloading the page)

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, merging this PR

PR #785, opened 2025-01-15
Fixes #26182: Datasources tests fail in 8.2 since restextractorservice deletion

clarktsiory:
https://issues.rudder.io/issues/26182

It is weird that we did not get an upmerge conflict from https://github.com/Normation/rudder-plugins/pull/784/files#diff-989df757ea97a81756f5beb30e2373c1467346fa13c2847917f2db0e54a284dcL68


fanf (approved):

fanf:
The git part of the upmerge is ok, it's a scala compilation problem. We don't rebuild everything on upmerge, it's too long. We just check that the merge is fast forward from a git point of view.

PR #787, opened 2025-01-16
Fixes #26199: Janino dependency was not provided to plugins causing plugin load issue

clarktsiory:
https://issues.rudder.io/issues/26199

Following  https://github.com/Normation/rudder-plugins/pull/786 : 
in fact `janino` has been the only dependency with a signed jar that was added in 8.3, and not provided to plugins, so it was causing the error.


clarktsiory:
Commit modified

fanf (approved):

PR #793, opened 2025-02-07
Fixes #26309: Datasources HTTP tests are failing in 8.2

clarktsiory:
https://issues.rudder.io/issues/26309

Just skipping them as are they are now rarely modified, at some point we will want to rewrite the whole testing framework, or change the setup to ignore some tests depending on the environment (like postgresql running only on CI)


clarktsiory:
Commit modified

amousset (approved):

PR #807, opened 2025-03-19
Fixes #26511: Branding plugin display is broken

RaphaelGauthier:
https://issues.rudder.io/issues/26511


clarktsiory (changes requested):
The media.css file has some still useful CSS : without it the preview will not display

image


Also the cards still look broken (somthing like `card-body p-0` is resetting the default padding within the content of the card, we may need to leave it as non-zero)

RaphaelGauthier:
PR updated with a new commit

RaphaelGauthier:
> The media.css file has some still useful CSS : without it the preview will not display
My bad, I renamed the file without using git, and forgot to upload the new .scss file..

RaphaelGauthier:
> Also the cards still look broken (somthing like `card-body p-0` is resetting the default padding within the content of the card, we may need to leave it as non-zero)

I don't understand or can't reproduce this feedback, with the right css file, the cards are displayed correctly, aren't they?
branding

clarktsiory (approved):
LGTM ! it works now !

fanf:
OK, squash merging this PR

PR #808, opened 2025-03-20
Fixes #26583: Navbar menu is broken

RaphaelGauthier:
https://issues.rudder.io/issues/26583

**Before:**
ui-changevalidation-old

**After:**
ui-changevalidation

### DIFF
--- a/change-validation/src/main/resources/template/ChangeValidationManagement.html
+++ b/change-validation/src/main/resources/template/ChangeValidationManagement.html
@@ -1,320 +1,313 @@
 <xml:group>
   <component-body>
 
-<head_merge>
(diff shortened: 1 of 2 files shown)


clarktsiory (commented):
:+1: LGTM, the menu is appearing and it fixes previous display issues

But the menu navigation is broken (which is not so problematic, we can fix it later), one annoying thing is that the page is cut before the end, the scroll is not reaching the bottom :
image
  
If you think it cannot be fixed easily, this PR can be merged and we can always do it later

fanf:
I'm for merging this one. We're just too near to the RC to wait on non blocking errors.

fanf (approved):

fanf:
OK, merging this PR

PR #810, opened 2025-03-24
Fixes #26608: CurrentUser.queryContext is null when used in a ZIO for (public plugins)

fanf:
https://issues.rudder.io/issues/26608

See https://github.com/Normation/rudder/pull/6280 for context


fanf:
OK, merging this PR

PR #816, opened 2025-04-15
Fixes #26752: Test with specific supervised target file permissions are failing on CI

clarktsiory:
https://issues.rudder.io/issues/26752

similar to https://github.com/Normation/rudder-plugins/pull/793/files, due to ns-remap on CI


clarktsiory:
OK, merging this PR

PR #817, opened 2025-04-15
Fixes #26756: Skipped tests are not skipped

clarktsiory:
https://issues.rudder.io/issues/26756

When `skipped` is put at the end, some assertions may still fail and throw before the `Skipped` exception is thrown


amousset (approved):

clarktsiory:
OK, merging this PR

PR #818, opened 2025-04-16
Update DataSourcesPluginDef.scala to fix URL

lordtaki:
In Plugins / Plugin Information, the link points to rudder/secure/administration/dataSourceManagement but page was available at rudder/secure/plugins/dataSourceManagement. I choosed to fix it based on a href reference.


clarktsiory:
That's right, there is a broken link to the plugin page !
Thanks for pointing that out !

However, we will try to avoid changing the URL of the page and the menu. We can keep `rudder/secure/plugins/dataSourceManagement`, and fix the link instead.

clarktsiory:
This is superseded by https://github.com/Normation/rudder-plugins/pull/820

PR #819, opened 2025-04-16
Fixes #26760: Add a new internal endpoint in order to replace WorkflowInformation

skaerg:
https://issues.rudder.io/issues/26760

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala
@@ -198,6 +198,31 @@ object ChangeRequestJson {
   }
 }
 
+/**
+ * Class that represents the number of change requests that are currently in a "pending" status, i.e.
+ * "Pending validation" and "Pending deployment" respectively.
+ * Both fields are optional : either field will be present if and only if the user who made the request has
+ * the required authorization type, i.e. Validator.Read, and Deployer.Read authorizations respectively.
+ *
+ * @param pendingValidation the current number of change requests that have the "Pending validation" status
+ * @param pendingDeployment the current number of change requests that have the "Pending deployment" status
+ */
+final case class PendingCountJson(
+    pendingValidation: Option[Long],
+    pendingDeployment: Option[Long]
+)
+
+object PendingCountJson {
+  implicit val encoder: JsonEncoder[PendingCountJson] = DeriveJsonEncoder.gen[PendingCountJson]
+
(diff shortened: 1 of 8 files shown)


skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

skaerg:
Commit modified

change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJson.scala
```diff
@@ -198,6 +198,22 @@ object ChangeRequestJson {
...
+object PendingCountJson {
+  implicit val encoder: JsonEncoder[PendingCountJson] = DeriveJsonEncoder.gen[PendingCountJson]
+
+  def from(map: Map[WorkflowNodeId, Long]): PendingCountJson = {
+    PendingCountJson(
+      map.get(TwoValidationStepsWorkflowServiceImpl.Validation.id),
+      map.get(TwoValidationStepsWorkflowServiceImpl.Deployment.id)
+    )
+  }
+}
```
clarktsiory, line 215: The case class `PendingCountJson` should have a documentation of why it has this structure, and why the `Option`, what it means and how it is supposed to be used.

Here, since the semantics of `None` and `Some(0)` are very different, there should be a warning about what the `None` is supposed to mean : it means that there are no rights.
It could as well have been another data type than `Option`, with the equivalent of `None` being named `NoRights`, but using an `Option` is also fine as long as it is well documented and has a very little scope...

change-validation/src/main/scala/com/normation/plugins/changevalidation/api/WorkflowInternalApi.scala
```diff
@@ -0,0 +1,105 @@
+package com.normation.plugins.changevalidation.api
```
clarktsiory, line 1: The copyright header is missing in this file

change-validation/src/test/scala/com/normation/plugins/changevalidation/api/WorkflowInternalApiTest.scala
```diff
@@ -0,0 +1,85 @@
+package com.normation.plugins.changevalidation.api
```
clarktsiory, line 1: The copyright header is missing in this file

clarktsiory (changes requested):
LGTM ! Nice change, there is just some missing comments : scaladoc and copyright

PR #822, opened 2025-04-22
Fixes #26770: WoChangeRequestJdbcRepository#updateChangeRequest executes several transactions

skaerg:
https://issues.rudder.io/issues/26770

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJdbcRepository.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJdbcRepository.scala
@@ -38,6 +38,9 @@
 package com.normation.plugins.changevalidation
 
 import cats.data.NonEmptyList
+import cats.syntax.applicative.*
+import cats.syntax.applicativeError.*
+import cats.syntax.functor.*
 import cats.syntax.reducible.*
 import com.normation.errors.*
 import com.normation.eventlog.EventActor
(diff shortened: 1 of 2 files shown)


skaerg:
Commit modified

skaerg:
Commit modified

change-validation/src/main/scala/com/normation/plugins/changevalidation/ChangeRequestJdbcRepository.scala
```diff
@@ -297,29 +299,31 @@ class WoChangeRequestJdbcRepository(
...
-          updateChangeRequestSQL(name, desc, xml, modId, changeRequest.id).run.transact(xa)
-        )
+    val process = {
+      for {
+        exists <- getSQL(changeRequest.id).option
+        _      <- exists match {
+                    case None    =>
+                      val msg =
+                        s"Change Request with id ${changeRequest.id.value} was not found in database"
+                      new Exception(msg).raiseError[ConnectionIO, Throwable]
```
clarktsiory, line 313: The exception type `IllegalArgumentException` may be more suited here, errors should better be informative.
We wouldn't bother to create a custom exception type, as it will only be used as a message container (but we would need a custom type and match on each one if we want to handle each database error specifically).

clarktsiory (changes requested):
There are some needed changes for Worfklow repository because the implementation semantics should not change : the update in db must not throw

skaerg:
Commit modified

skaerg:
Commit modified

clarktsiory (approved):
LGTM, the logs are now standardized, and outside of the `ConnectionIO` operations.
And having a single transaction which checks the existence of a change requests and updates it guarantees that it will still be there

PR #832, opened 2025-05-15
Fixes #26906: Migrate the ChangeValidationSettings snippet from Scala/lift to Elm

skaerg:
https://issues.rudder.io/issues/26906

This PR is part of the plan to remove the dependency to lift from change-validation.
The goal of this PR is to port the `Change validation status` section of the change-validation settings page to Elm, thereby replacing the `ChangeValidationSettings` lift snippet. 

This new section was integrated with the existing Elm app that renders the 
`SupervisedTargets` and `WorkflowUsers` sections, which were previously two separate Elm apps (see https://github.com/Normation/rudder-plugins/pull/840). The entire change-validation settings page is now rendered by a single Elm app that contains all three sections.

image

Notes : 
- The api calls that were previously executed by one or more separate Elm apps or snippets are now called only once (e.g. the initial `GET` api call in order to display the current settings' values) and handled separately by each section.
- The Elm app is loaded if and only if the change-validation settings tab is clicked/loaded (it used to be loaded even if the user only visited the other settings pages, e.g. the `General settings` page)
- Some info-sections were rephrased for clarity and/or modified when there were contradictions between two different info-sections (see https://github.com/Normation/rudder-plugins/pull/832/commits/31ed45f4680680ca119c7db5dfe5f3bcf7890b13 )
- The checkboxes now use bootstrap 5 as well as fontawesome icons, thereby replacing the uses of `ion ion-checkmark` in this page


skaerg:
PR updated with a new commit

skaerg:
Commit modified

skaerg:
Commit modified

change-validation/src/main/elm/sources/ChangeValidationSettings.elm
```diff
@@ -21,7 +26,14 @@ init flags =
...
             initModel flags.contextPath flags.hasWriteRights
     in
-    ( m, Cmd.batch [ getUsers m.workflowUsersModel, getValidateAllSetting m.workflowUsersModel, getTargets m.supervisedTargetsModel ] )
+    ( m
+    , Cmd.batch
+        [ getChangeValidationStatus m.workflowSettingsModel
+        , getAllWorkflowSettings m
+        , getUsers m.workflowUsersModel
+        , getTargets m.supervisedTargetsModel
+        ]
```
clarktsiory, line 35: Fine, but the `getChangeValidationStatus` call is not necessary because the Elm app is only loaded when the plugin is enabled anyway...
(and if we really had to check this, it would not be all commands in batch in the `init`, but rather getting the status first, then getting all other data in batch)

change-validation/src/main/elm/sources/WorkflowSettings.elm
```diff
@@ -35,17 +53,116 @@ update msg model =
...
+        settingName =
+            case setting of
+                WorkflowEnabled ->
+                    "enable_change_request"
+
+                SelfValidation ->
+                    "enable_self_validation"
+
+                SelfDeployment ->
+                    "enable_self_deployment"
```
clarktsiory, line 106: this is used more than once across the different functions : it should be refactored as 
```elm
settingText : WorkflowSetting -> String
```

clarktsiory:
yes you can do improvements on the HTML content, it was in the HTML template but now it's in an Elm file so if you don't do it in Elm, it will not be fixed (if we also fix it in previous version with HTML, it will need to be fixed by hand in the Elm file during upmerge)

clarktsiory:
I don't know which one is correct, so it needs to be tested

change-validation/src/main/elm/sources/DataTypes.elm
```diff
@@ -84,9 +89,103 @@ type alias FormState =
...
+------------------------------
+-- WORKFLOW SETTINGS --
+------------------------------
+
+
+type alias WorkflowSettingsModel =
+    { contextPath : String
+    , canWrite : Bool
+    , viewState : ViewState
+    }
```
clarktsiory, line 102: This belongs to the `WorkflowSettings` module

clarktsiory: `canWrite` does not follow our convention so it makes it hard to find and the specificity will lead to believe that is an exception and behaves differently
```suggestion
type alias WorkflowSettingsModel =
    { contextPath : String
    , hasWriteRights : Bool
    , viewState : ViewState
    }
```

change-validation/src/main/elm/sources/WorkflowSettings.elm
```diff
@@ -0,0 +1,380 @@
...
+            model |> updateSetting result SelfDeployment
+
+        SwapWorkflowEnabled ->
+            ( model |> swapSettingOn WorkflowEnabled, Cmd.none )
+
+        SwapSelfValidation ->
+            ( model |> swapSettingOn SelfValidation, Cmd.none )
+
+        SwapSelfDeployment ->
+            ( model |> swapSettingOn SelfDeployment, Cmd.none )
```
clarktsiory, line 67: So, there already is the type 
```elm
type WorkflowSetting
    = WorkflowEnabled
    | SelfValidation
    | SelfDeployment
```
So why not use them in the messages ?

It could be simplified to : 
```elm
type Msg =
   ...
   | SaveWorkflowSetting WorkflowSetting (Result Err Bool)
   | ToggleSetting WorkflowSetting
```
And in the update 
```suggestion
        SaveWorkflowSetting setting result ->
           updateSetting setting result model

        ToggleSetting ->
            ( model |> toggleSettingOn SelfDeployment, Cmd.none )
```

change-validation/src/main/elm/sources/WorkflowSettings.elm
```diff
@@ -0,0 +1,380 @@
...
+                        SelfDeployment ->
+                            { formSettings | selfDeployment = newValue }
+            in
+            { model | viewState = WorkflowSettingsView { initSettings = newSettings, formSettings = newSettings } }
+
+
+{-| Swap the current value of a setting in the view (when a given setting is clicked)
+-}
+swapSettingOn : WorkflowSetting -> WorkflowSettingsModel -> WorkflowSettingsModel
+swapSettingOn setting model =
```
clarktsiory, line 135: `toggle` is more suited

change-validation/src/main/elm/sources/WorkflowSettings.elm
```diff
@@ -0,0 +1,380 @@
...
+    case model.viewState of
+        InitWorkflowSettingsView ->
+            model
+
+        WorkflowSettingsView form ->
+            let
+                editedForm =
+                    swapSettingOnForm setting form.formSettings
+            in
+            { model | viewState = WorkflowSettingsView { form | formSettings = editedForm } }
```
clarktsiory, line 157: to save some vertical space and for readability  
```suggestion
        swapSettingOnForm f =
            case setting of
                WorkflowEnabled ->
                    { f | workflowEnabled = not f.workflowEnabled }

                SelfValidation ->
                    { f | selfValidation = not f.selfValidation }

                SelfDeployment ->
                    { f | selfDeployment = not f.selfDeployment }
    in
    case model.viewState of
        InitWorkflowSettingsView ->
            model

        WorkflowSettingsView form ->
            { model | viewState = WorkflowSettingsView { form | formSettings = swapSettingOnForm form.formSettings } }
```

change-validation/src/main/elm/sources/DataTypes.elm
```diff
@@ -95,7 +194,7 @@ type WorkflowUsersMsg
       GetUsers (Result Error UserList)
     | RemoveUser (Result Error Username)
     | SaveWorkflow (Result Error UserList)
-    | CallApi (Model -> Cmd Msg)
+    | CallApi (WorkflowUsersModel -> Cmd Msg)
```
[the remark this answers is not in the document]
clarktsiory: it's explained in another comment (https://github.com/Normation/rudder-plugins/pull/832#discussion_r2131808097), the `WorkflowUsers` module should not know about the parent `Msg` in the composite pattern, or else it defeats the purpose of encapsulating the child.
A solution could be to avoid having a generic  `CallApi` but use concrete messages for API calls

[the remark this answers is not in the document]
skaerg: Regardless of the `CallApi` type, each sub-section of the main app (`SupervisedTargets`, `WorfklowUsers`, `WorkflowSettings`) depends on the Msg type in order to display the app in their individual view functions : the main app has a view function that returns a `Html Msg` and the display of each sub-section is delegated to its dedicated module, hence why the view function of each subsection also returns a `Html Msg`.

Is there a way to display each section without depending on the `Msg` type  (i.e. is there a way to "convert" a `Html WorkflowUsersMsg` into a `Html Msg`, and so on) ?

change-validation/src/main/elm/sources/SupervisedTargets.elm
```diff
@@ -1,6 +1,5 @@
-module SupervisedTargets exposing (alphanumericRegex, decodeApiCategory, decodeApiSave, decodeCategory, decodeSubcategories, decodeTarget, displayCategory, displaySubcategories, displayTarget, encodeTargets, getSupervisedIds, getTargets, initModel, isAlphanumeric, saveTargets, update, updateTarget, view)
+module SupervisedTargets exposing (Category, Subcategories, SupervisedTargetsModel, SupervisedTargetsMsg, Target, getTargets, initModel, update, view)
```
clarktsiory, line 1: `getTargets` could be removed from the publicly exposed API, we don't need to expose this specific API call : what the parent app needs to know is what "initial commands" this component needs, so you can define  
```elm
initCmd: Cmd SupervisedTargetsMsg
initCmd =
    getTargets parentModel

getTargets : { a | contextPath : ContextPath } -> Cmd SupervisedTargetsMsg
getTargets { contextPath } =
    ...
```

change-validation/src/main/elm/sources/WorkflowSettings.elm
```diff
@@ -1,16 +1,61 @@
-module WorkflowSettings exposing (..)
+module WorkflowSettings exposing (ViewState, WorkflowSettingsModel, WorkflowSettingsMsg, initModel, initWorkflowSettingsView, update, view)
```
clarktsiory, line 1: it seems weird that multiple `init` are exposed, the second one is not really and `init`, it's a setter, that has a confusing name.

I see that it is used in the parent with : 
```
                      { model
                        | workflowSettingsModel = { wsModel | viewState = WorkflowSettings.initWorkflowSettingsView settings }
                        ...
                      }
```

But the parent does not need to know that "the `viewState` field of the child model needs to be updated !".
What it wants is only an `updateSetting`/`setSetting` directly on the child model, so you could change the function to directly modify the `WorkflowSettingsModel`, I guess it would be an `update` then

change-validation/src/main/elm/sources/WorkflowUsers.elm
```diff
@@ -1,15 +1,17 @@
-module WorkflowUsers exposing (..)
+module WorkflowUsers exposing (EditMod, ValidateAllView, WorkflowUsersModel, WorkflowUsersMsg, WorkflowUsersView, getUsers, initModel, initValidateAllForm, update, view)
```
clarktsiory, line 1: same as other comments on both concerns : 
* `getUsers` is part of initial commands
* `initValidateAllForm` probably has the wrong name and signature

clarktsiory (changes requested):
Very nice result, it's relieving to finally have encapsulation in modules :tada: 
There are still some minor changes

change-validation/src/main/elm/sources/WorkflowUsers.elm
```diff
@@ -282,24 +415,575 @@ mapUserList field f viewState =
...
+                [ label [ style "vertical-align" "middle", style "display" "inline-block" ]
+                    [ if isEditActivate then
+                        input
+                            [ class "box-input-element center-box-element"
+                            , type_ "checkbox"
+                            , value user.username
+                            , checked isChecked
+                            ]
+                            []
+
```
skaerg, line 710: The checkboxes already seem to be synchronized with the model

PR #837, opened 2025-05-23
Fixes #26951: Plugins need CSP to be strict in Rudder but disabled in plugin pages

clarktsiory:
https://issues.rudder.io/issues/26951


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

PR #840, opened 2025-05-27
Fixes #26965: Merge the SupervisedTargets and WorkflowUser Elm apps in a single Elm app

skaerg:
https://issues.rudder.io/issues/26965

This is part of the plan to migrate the change-validation snippets from Scala to Elm.
This PR aims to "merge" the `SupervisedTargets` and `WorkflowUsers` Elm apps of the change-validation settings page as one app.

Commit history : 
1. Modify the structure of the html page so that the SupervisedTargets and WorkflowUsers apps are side by side
2. Merge both apps as one model (composite pattern) : the main app (ChangeValidationSettings) receives the users' input, and its handling is delegated to the appropriate part of the page (either WorkflowUsers or SupervisedTargets)
3. Create a separate viewState for the workflowUsers app (similar to the existing viewState for the validate-all checkbox)
4. Refactor wokflowUsers view according to new viewstate
5. fix style issues ; fix validate all button (it used to reload the whole page when clicked)
6. replace .ion checkboxes with fontawesome checkboxes in the supervisedtargets section
7. review corrections


change-validation/src/main/elm/sources/WorkflowUsers.elm
```diff
@@ -177,33 +235,124 @@ update msg model =
...
+                    WorkflowUsers { formState | validatedUsers = f formState }
+
+                UnvalidatedUsers ->
+                    WorkflowUsers { formState | unvalidatedUsers = f formState }
+
+                RightChecked ->
+                    WorkflowUsers { formState | rightChecked = f formState }
+
+                LeftChecked ->
+                    WorkflowUsers { formState | leftChecked = f formState }
```
clarktsiory, line 302: those pattern match are the only one being used, the method can be simplified (and its params)

change-validation/src/main/elm/sources/WorkflowUsers.elm
```diff
@@ -21,137 +17,191 @@ filterUnvalidatedUsers users =
...
-                    ( { model | users = filter (\m -> m.username /= removeUser) model.users }, Cmd.none )
+                    ( { model | workflowUsersView = setUserList3 (filter (\m -> m.username /= removeUser)) Users model.workflowUsersView }, Cmd.none )
 
                 Err error ->
-                    ( model, errorNotification ("An error occurred while trying to delete a validated users:" ++ getErrorMessage error) )
+                    let
+                        errMsg =
+                            "An error occurred while trying to delete a validated users:" ++ getErrorMessage error
+                    in
+                    ( workflowUsersError model errMsg, errorNotification errMsg )
```
clarktsiory, line 62: the model should be left as is, the error should not impact the whole UX, a notification will be sufficient

change-validation/src/main/elm/sources/WorkflowUsers.elm
```diff
@@ -177,33 +205,106 @@ update msg model =
...
+                        |> setUserList formState.validatedUsers LeftChecked
+                        |> setUserList [] RightChecked
+
+                Right ->
+                    viewState
+                        |> setUserList formState.unvalidatedUsers RightChecked
+                        |> setUserList [] LeftChecked
+
+        _ ->
+            viewState
```
clarktsiory, line 260: I expect a setter of user list on a field with the type `UserList`, but it could be made a lot clearer and taken advantage by changing it a bit : 
* this is not an actual setter in the sense of optics/lenses in functional programming, but it is rather a "partial" setter, which is only a setter when applied to the concrete field, which is `UserListField` here, so it should be named differently and have the following signature or else it will be breaking conventions : `setUserListOn : UserListField -> UserList -> ... -> WorkflowUsersView`
* it is a first step low-level abstraction over `{ model | myField = setOn MyField }`, but it has no significant advantage over defining the boilerplate of individual field with their respective setters : `setMyField : Fieldtype -> Model -> Model`
* the advantage could be in the higher-level abstractions that use it, in practice we can abstract over known operations on the model, for instance : 
  * `checkAllView` is a good example
  * but in general `LeftChecked` and `RightChecked` seem to be updated together often, so why not defining a `setChecked left right = setUserListOn LeftChecked left >> setUserListOn RightChecked right` ? 
  * it is the same case for `ValidatedUsers` and `UnvalidatedUsers`
  * I even see higher-level operations such as `moveUsersToLeft`, `clearChecked`, etc.

So, the function needs to be renamed, and it's already a good start for abstracting over operations on the model, but this necessitates taking the time to understand what actually happening on the model before doing all the abstractions (the best timing for doing that is when creating the model, applying changes, adding operations, or refactoring the logic, but this PR does not fall into that category, so the renaming should be enough for now)

change-validation/src/main/elm/sources/WorkflowUsers.elm
```diff
@@ -177,33 +205,106 @@ update msg model =
...
+
+
+mapUserList : (UserList -> UserList) -> UserListField -> WorkflowUsersView -> WorkflowUsersView
+mapUserList f field viewState =
```
clarktsiory, line 264:
```suggestion
updateUserListOn :  UserListField -> (UserList -> UserList) -> WorkflowUsersView -> WorkflowUsersView
mapUserListOn field f viewState =
```

change-validation/src/main/resources/template/ChangeValidationManagement.html
```diff
@@ -181,72 +179,10 @@ <h3 class="page-title">Configure change request triggers</h3>
...
-						Configuring groups below will hence have no effect on validated users (in the list above), but will apply to non-validated users, who will still need a change request to modify a node from a supervised group.
-					  </p>
-					</div>
-				  </div>
-				</div>
+			  <div id="changeValidationSettings"></div>
 
-			  </div>
-			  </lift:authz>
+			  <!--
```
clarktsiory, line 184: comments of HTML code should be removed, they are not supposed to be sent in the HTML page being served, or it will reduce bandwidth...

change-validation/src/main/resources/template/ChangeValidationManagement.html
```diff
@@ -317,7 +255,35 @@ <h3 class="page-subtitle">Configure groups with change validations</h3>
...
+				</lift:authz>
+
+				<script>
+					$(document).ready(function () {
+                      var cvs = document.getElementById("changeValidationSettings");
+                      var initValues = {
+                        contextPath: contextPath,
+                        hasWriteRights: hasWriteRights
+                      };
+                      var app = Elm.ChangeValidationSettings.init({node: cvs, flags: initValues});
```
clarktsiory, line 276: use of `const` is unambiguous and should always be preferred over `var`

clarktsiory:
:+1: it just need a rebase and squash into a single commit before merging

PR #842, opened 2025-06-03
Fixes #27029: Change request count Elm WorkflowInformation app needs a loader  

clarktsiory:
https://issues.rudder.io/issues/27029
Peek 2025-06-03 15-45

There is some refactoring that justify that we need to add a loader :
* we need to change the initialization of the Elm app to be in an AJAX, instead of in a `<script/>` using the actor's `partialUpdate` (as in https://github.com/Normation/rudder/pull/5966), because script tags are annoying within actors and when we add CSP headers (see https://github.com/Normation/rudder-plugins/pull/837)
* the AJAX is a bit slower which makes it noticeable, so we need to add placeholders for display, [using bootstrap classes](https://getbootstrap.com/docs/5.1/components/placeholders/)
* the access checks, when moved to the snippet of the plugin extension `TopBarExtension`, simplifies the code in the actor a lot

### DIFF
--- a/change-validation/src/main/elm/sources/WorkflowInformation.elm
+++ b/change-validation/src/main/elm/sources/WorkflowInformation.elm
@@ -222,32 +222,27 @@ update msg model =
 
 view : Model -> Html Msg
 view model =
-    case model.pendingCount of
-        PendingCountWithTotal pc ->
-            li
-                [ Attr.class "nav-item dropdown notifications-menu"
-                , Attr.id "workflow-app"
-                ]
-                [ a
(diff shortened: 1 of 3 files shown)


fanf:
Works well, but I find extremely annoying to have the loader moving the frame at each page change. It would be much less unsetling if the width was fixed, and the loading placeholder didn't change it - ie, if only the digit(s) were replaced by something, not the whole bubble.

fanf:
Created sub-issue https://issues.rudder.io/issues/27033 to track the non-static width thingy.

PR #843, opened 2025-06-03
Fixes #27032: Update plugins parent pom to use scala 3

fanf:
https://issues.rudder.io/issues/27032

Almost went smootly. We get the usual syntatic changes, and a couple of less nice ones. 

# Plugin dependencies / pom.xml

This went better than I thought, with some "I have no idea how maven resolves things at all". 

- I chose to use the numbered version of scala binary directly in the package name. Using `_${scala-binary-version}` is getting really confusing given we have two now, and it's not like we are changing again in near future. So we now have `_3` and `_2.13` in package names. 
- I added the second scala lib (one for each version). This mimics the main pom, I'm not absolutly sure we need it repeatead, but we are sure we need it, so... 
- we find back changes from core dep:
    - `sealarate` is removed
    - `cron4s` get the `atto` backend to avoid pb with scala parser used in liftweb
-  `cats-effect-std` is needed for tests using something related to database (doobie indrirectly), I made it provided at that level

# Syntatic and trivial changes

- some type asciption as always
- removing final for objects
- adding `then` part on if
- removing unused variables in pattern matching/for loops
- renaming shadowed variables
- `withCoproductInstance` => `withSealedSubtypeHandled`

# Better type resolution

There is several place where the type resolution is better and we are forced to be more sincere about what we do. 
For example, in `datasources/src/test/scala/com/normation/plugins/datasources/UpdateHttpDatasetTest.scala` there is several places where we were throwing exception in place of dealing with error. This need to be a real `die()` now. 

# ChangeValidation: Doobie implicits and `extends AnyVal`

As in https://github.com/Normation/rudder/pull/6369 we have some `case class ... extends AnyVal` that is not supported anymore. We need to implement the related codec and base check. 
Identified case are: 
- `WorkflowNodeId`
- `ChangeRequestId`

# Datasource: invalidate state in macro with `PatcherConfiguration`

I had an error about invalidate state for `patchCfg`, and not understanding it at all. It seems that the macro resolution does not support that anymore. 
Resolved by having the configuration `.ignoreNoneInPatch` used at each call site.

### DIFF
--- a/api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/ApiAccountsExtension.scala
+++ b/api-authorizations/src/main/scala/com/normation/plugins/apiauthorizations/ApiAccountsExtension.scala
@@ -75,7 +75,7 @@ class ApiAccountsExtension(val status: PluginStatus)(implicit val ttag: ClassTag
    *   ...
    * ]
    */
-  def render(xml: NodeSeq) = {
+  def render(xml: NodeSeq): NodeSeq = {
     // get all apis and for public one, and create the structure
     import net.liftweb.http.js.JsCmds.*
     import net.liftweb.http.js.JE.*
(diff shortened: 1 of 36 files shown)


fanf:
OK, merging this PR

PR #844, opened 2025-06-05
Fixes #27000: Datasources edition has action buttons even for read-only user

clarktsiory:
https://issues.rudder.io/issues/27000

* Action buttons were missing the `hasWriteRights` check
* `administration_write` permission allows to edit datasources, not `rule_write`


clarktsiory:
OK, merging this PR

PR #847, opened 2025-06-11
Fixes #26908: Migrate the ChangeRequestChangesForm snippet from Scala/lift to Elm

skaerg:
https://issues.rudder.io/issues/26908


change-validation/src/main/scala/com/normation/plugins/changevalidation/api/WorkflowInternalApi.scala
```diff
@@ -83,20 +102,60 @@ object WorkflowInternalApi       extends Enum[WorkflowInternalApi] with ApiModul
     }
   }
 
+  case object ChangeRequestDetailsWithHistory extends WorkflowInternalApi with OneParam with StartsAtVersion21 with SortIndex {
+    val z              = implicitly[Line].value
+    val (action, path) = GET / "changevalidation" / "workflow" / "changeRequestWithHistory" / "{id}"
+    val description    =
+      "Get the main details and list of logs of a change request by its ID."
```
clarktsiory, line 109: With the "main details and list of logs" it also returns all that is needed to display the details of the page eagerly (among which, the tree of changes).

I suggest `ChangeRequestMainDetails`

clarktsiory (changes requested):

change-validation/src/main/scala/com/normation/plugins/changevalidation/api/WorkflowInternalApi.scala
```diff
@@ -139,4 +198,188 @@ class WorkflowInternalApiImpl(
...
+
+  object ChangeRequestChanges extends LiftApiModule {
+    override val schema:  WorkflowInternalApi.ChangeRequestChanges.type = API.ChangeRequestChanges
+    implicit val encoder: JsonEncoder[ChangeRequestChangesJson]         = ChangeRequestChangesJson.encoder
+
+    // Checks if external validation is needed
+    def checkWorkflow: Boolean = workflowService.needExternalValidation()
+
+    private[this] def disabledWorkflowAnswer[T]: IOResult[T] = {
+      Inconsistency("Workflows are disabled in Rudder, the internal workflow API is not available").fail
```
clarktsiory, line 268: you could define these in a shared scope (in the parent class `WorkflowInternalApiImpl`) instead of copying it thrice).

I assume it's the same thing with `withChangeRequestContext`, and maybe other methods that are copied across endpoint implementations

clarktsiory (approved):
I fixed tests, and added some linter rules, and I made the elm-git-install run consistently

PR #851, opened 2025-06-12
Fixes #27043: Change request details history is not updated when updating via API

skaerg:
https://issues.rudder.io/issues/27043


change-validation/src/main/scala/com/normation/plugins/changevalidation/api/ChangeRequestApi.scala
```diff
@@ -410,12 +411,18 @@ class ChangeRequestApiImpl(
...
-          for {
-            updated    <- writeChangeRequest.updateChangeRequest(newCR, authzToken.qc.actor, None)
-            serialized <- serialize(updated, status).toIO
-          } yield {
-            serialized
+          workflowLevelService.getWorkflowService() match {
+            case ws: TwoValidationStepsWorkflowServiceImpl =>
+              for {
+                newCR      <- ws.updateChangeRequestInfo(changeRequest, newInfo, userService.getCurrentUser.actor, None)
+                updated    <- writeChangeRequest.updateChangeRequest(newCR, authzToken.qc.actor, None)
```
clarktsiory, line 418: the `writeChangeRequest.updateChangeRequest` seems to be redundant, the workflow level service already updates the change request repository

PR #857, opened 2025-06-17
Fixes #27115: Add a cache for OIDC validation request

fanf:
https://issues.rudder.io/issues/27115


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
```diff
@@ -1367,15 +1375,82 @@ object RudderOpaqueTokenAuthenticationProvider {
...
-    val a = opaqueTokenAuthenticationProvider.authenticate(authentication)
-    a
+    authentication match {
+      case token: BearerTokenAuthenticationToken =>
+        // check is that token already have an authentication
+        cache match {
+          // cache defined:
+          case Some(c) =>
+            c.get(token.getToken, (_: String) => doRemoteAuthentication(token)) match {
+              case Left(ex)                                      => throw ex
```
clarktsiory, line 1433: is seems like `Left` values errors are always stored in cache. Is it not potentially an issue if the called is filled up with wrong tokens even so that valid tokens are discarded ?
Why wouldn't we want to simply return the error ?

There seems to be solutions to discard the value in the cache : like returning null or so...
But it's really a Java API, we could use https://github.com/blemale/scaffeine

auth-backends/src/test/scala/com/normation/plugins/authbackends/TestCache.scala
```diff
@@ -0,0 +1,218 @@
...
+      introspector.count.get === 1
+    }
+
+    "second request ko even if in cache" in {
+      c = 5
+      val res = Try(providerCache.authenticate(GOOD_TOKEN_EXP1.toAuth))
+
+      res must beAnInstanceOf[scala.util.Failure[?]]
+      introspector.count.get === 1
+    }
```
clarktsiory, line 216: It seems that the failure should contain an `InvalidBearerTokenException` as required by Spring, shouldn't the exception type be explicit in the test ?

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -268,6 +274,7 @@ object RudderRegistrationPropertyCommon {
     A_URI_LOGOUT_REDIRECT     -> "(optional) the redirect URL to provide to the IdP after logout",
     A_URI_INTROSPECT          -> "in case of opaque access bearer token, the introspect URL on which the token must be validated",
     A_PIVOT_ATTRIBUTE         -> "the attribute used to find local app user (OIDC user authentication) or the local API token (opaque bearer token)",
+    A_CACHE_VALIDATION        -> "(optional) the duration (number followed by a time unit) for which an opaque token validation should be cache",
```
clarktsiory, line 277:
```suggestion
    A_CACHE_VALIDATION        -> "(optional) the duration (number followed by a time unit) for which an  validated opaque token should be cached",
```

clarktsiory (changes requested):
[The spec says](https://datatracker.ietf.org/doc/html/rfc7662#section-4) : 
> If the response contains the "exp" parameter (expiration), the response MUST NOT be cached beyond the time indicated therein.

I don't know where that condition could be checked then...

Otherwise I had some minor remarks and suggestions

auth-backends/src/test/scala/com/normation/plugins/authbackends/TestCache.scala
fanf: I don't see why we would limit to that. It could be any `AuthenticationFailure`. 
But even then, I think we really want to catch any exection raised by the underlying framework, and spring will do the translation reagarding resource cleaning accordingly.

fanf:
PR updated with a new commit

PR #860, opened 2025-06-23
Fixes #27149: Enable spotless for plugins

fanf:
https://issues.rudder.io/issues/27149

We are forced to duplicate in `<plugin-management>` and `<plugin>` the same things, I'm not sure why but let's say it's the way of maven. 

Also, add  `dependency-reduced-pom.xml` to git ignore. It looks like it's produced by modern maven to help reducing the number of dependencies, but it doesn't fit well with our parent pom bringing a lot of things.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ node_modules
 */*/*/gulpfile.mjs
 **/toserve/*/*.css
 **/toserve/*/*.scss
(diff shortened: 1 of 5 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory (approved):
:rocket: I pushed to fix the format for some plugins

PR #871, opened 2025-07-15
Fixes #27278: Post-snippet migration fixes and cleanup in change-validation

skaerg:
https://issues.rudder.io/issues/27278

This PR aims to mostly finalise the migration of the change-validation lift snippets that have been migrated to Elm this far, which includes: 
- a modification in the `ChangeRequestDetails` Elm app and in the `WorkflowInternalApi` : 
  - the `changeRequestMainDetails` endpoint now returns the list of all possible next statuses for a given change request, rather than up to a single possible status.
  - The Elm app now displays a `select` element so that the user can pick the next status of the change request in case there are several reachable ones - a "next" status is deemed reachable if the user has the the necessary authorization type in order to move the change request to that status, i.e. `validator_write` to move a change request to the `Pending deployment` status, and `deployer_write` to move a change request to the `Deployed` status.
  - The Elm app uses flags in order to determine the current user's rights.
- deleting the 4 lift snippets that have been migrated, i.e. 
  - ChangeValidationSettings.scala
  - ChangeRequestEditForm.scala
  - ChangeRequestChangesForm.scala
  - ChangeRequestDetails.scala


skaerg:
PR updated with a new commit

change-validation/src/main/elm/sources/ChangeRequestDetails.elm
```diff
@@ -721,6 +774,64 @@ stepChangeAction stepChange =
...
+            Next allNextSteps.selected
+
+
+statusSelect : AllStepChanges -> Html Msg
+statusSelect allStepChanges =
+    case allStepChanges of
+        BackSteps backStatus ->
+            span
+                [ class "well" ]
+                [ text ("  " ++ stepChangeToString (Back backStatus) ++ "  ") ]
```
clarktsiory, line 793:
```suggestion
                [ text <| String.Extra.surround "   " (stepChangeToString (Back backStatus)) ]
```

change-validation/src/main/elm/sources/ChangeRequestDetails.elm
```diff
@@ -721,6 +774,64 @@ stepChangeAction stepChange =
...
+                                    "Pending deployment" ->
+                                        succeed (ChangeSelectedNextStep PendingDeployment)
+
+                                    "Deployed" ->
+                                        succeed (ChangeSelectedNextStep Deployed)
+
+                                    _ ->
+                                        fail "Invalid status"
+                            )
+                    )
```
clarktsiory, line 830: `on "change"` works and has some low-level API (need to decode the "targetValue" of the event)
but I'm used to seeing `onInput` because it simplifies the API, but it always requires a message. Since we need to "decode" the string value, if it fails we could just emit an "Ignore" message

change-validation/src/main/scala/com/normation/plugins/changevalidation/api/WorkflowInternalApi.scala
```diff
@@ -210,6 +211,23 @@ class WorkflowInternalApiImpl(
...
+      def findAllNextStepsAux(curStatus: Option[WorkflowNodeId], acc: Seq[WorkflowNodeId]): Seq[WorkflowNodeId] = {
+        curStatus match {
+          case Some(cur) =>
+            val next = workflowService.findNextStatus(cur)
+            if (!acc.contains(next)) {
+              findAllNextStepsAux(next, acc :+ cur)
+            } else acc
+          case None      => acc
+        }
+      }
```
clarktsiory, line 226: 👍 but I wonder why the check is needed : `acc.contains(next)` ?

It seems that there is already an assumption in the business logic that the "next status" of a status cannot be the same status itself.

If that is the reason for the check, then the check should be at the level of the "business logic" itself e.g. in the implementation of `findNextStatus(nextStatus: Option[WorkflowNodeId]): Option[WorkflowNodeId]` if it really needs to be enforced, plus the appropriate unit tests for testing that it never happens. When doing that in a consumer of the business logic API, we don't want all consumers to do that check, it would not be the right architecture. So we could just assume that from the finite state machine.
TLDR : I think we wouldn't want that hassle here, just not doing the check is fine

change-validation/src/main/elm/sources/ChangeRequestDetails.elm
```diff
@@ -450,65 +484,81 @@ bannerView model cr =
...
+        isReachableStatus =
+            \status ->
+                ((status == PendingDeployment) && model.hasValidatorWriteRights)
+                    || ((status == Deployed) && model.hasDeployerWriteRights)
+
+        reachable =
+            allNextSteps
+                |> List.filter isReachableStatus
+                |> uniqueBy (Next >> stepChangeToString)
+
```
clarktsiory, line 532: This is in the _view_, the right place for assigning the `reachable` statuses is when getting the `allNextStatuses` from the API.

As an architecture rule, we don't do logic in the _view_, or else it would make our logic not testable (we test the _update_, not the _view_),  and we should design the model for that, and here the model is already fit for that : we would directly need 
```elm
type alias Model =
   {
    ...
    , allNextSteps : AllNextSteps
   }
```
instead of 
```elm
type alias Model =
    {
    ...
    , allNextSteps : List NextStatus
    }
```

clarktsiory (changes requested):
Nice !
Some review on the architecture, and some suggestion

skaerg:
PR updated with a new commit

clarktsiory (approved):
LGTM, congrats on that last bit of migration which is a great contribution for this plugin, especially in the Elm code base 👏

PR #872, opened 2025-07-16
Fixes #27289: Change request page is not updated directly after using the form to edit the title/description 

skaerg:
https://issues.rudder.io/issues/27289

Previously, the banner and changes table of a given change request were not updated directly after the user had used the edit form in order to modify the title or description of the change request.
The banner and changes table are now updated without needing to refresh the page.

Note : After editing the title or description of a change request, no new log will appear in the logs table - the issue is on the API side (modifying a change request's info by using the public API does not create a new log). Hence, the update of the changes table is not visible in this case.

### DIFF
--- a/change-validation/src/main/elm/sources/ChangeRequestDetails.elm
+++ b/change-validation/src/main/elm/sources/ChangeRequestDetails.elm
@@ -294,10 +294,15 @@ update msg model =
 
         EditFormMsg editFormMsg ->
             let
-                ( efModel, efCmd ) =
+                ( efModel, efCmd, efOutcome ) =
                     EditForm.update editFormMsg model.editForm
             in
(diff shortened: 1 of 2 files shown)


clarktsiory (changes requested):
The only source of truth is the data from the API, and the solution to have the most up-to-date information here is to call the API to get the last details of the change request.

After the update of the form, there should be a command to do that API call. This will automatically set the form details to the latest, and also it should refresh the "change history" table, which is supposed to have an additional entry "Change request changed" (but there is an ongoing issue with that https://github.com/Normation/rudder-plugins/pull/851)

skaerg:
PR updated with a new commit

clarktsiory (approved):
LGTM, now this applies a state-of-the-art Elm architecture 👍

PR #874, opened 2025-07-24
Fixes #27352: Switch back to alpha 1

amousset:
https://issues.rudder.io/issues/27352


VinceMacBuche:
Why switching back to alpha 1 ?  Plugins were all tagged and build no ?

PR #852, opened 2025-06-12
Fixes #27085: Change validation acceptation and decline endpoints fail with permission issues

clarktsiory:
https://issues.rudder.io/issues/27085

Use simpler authorization structures from https://github.com/Normation/rudder-plugins/pull/879 to instantiate implicits early, thanks to changing the interface in https://github.com/Normation/rudder/pull/6453.

PR #879, opened 2025-08-27
Fixes #27489: Impact of refactoring authorizations in plugins

clarktsiory:
https://issues.rudder.io/issues/27489

Impact of https://github.com/Normation/rudder/pull/6570

### DIFF
--- a/change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
+++ b/change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
@@ -333,12 +333,12 @@ class TwoValidationStepsWorkflowServiceImpl(
    */
   private def canValidate(isCreator: Boolean, selfValidation: () => IOResult[Boolean]): Boolean = {
     val correctActor = selfValidation().orElseSucceed(false).runNow || !isCreator
-    correctActor && userService.getCurrentUser.checkRights(AuthorizationType.Validator.Edit)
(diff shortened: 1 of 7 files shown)


change-validation/src/main/scala/com/normation/plugins/changevalidation/WorkflowService.scala
```diff
@@ -333,12 +333,12 @@ class TwoValidationStepsWorkflowServiceImpl(
...
   private def canValidate(isCreator: Boolean, selfValidation: () => IOResult[Boolean]): Boolean = {
     val correctActor = selfValidation().orElseSucceed(false).runNow || !isCreator
-    correctActor && userService.getCurrentUser.checkRights(AuthorizationType.Validator.Edit)
+    correctActor && userService.getCurrentUser.map(_.checkRights(AuthorizationType.Validator.Edit)).getOrElse(false)
   }
 
   private def canDeploy(isCreator: Boolean, selfDeployment: () => IOResult[Boolean]): Boolean = {
     val correctActor = selfDeployment().orElseSucceed(false).runNow || !isCreator
-    correctActor && userService.getCurrentUser.checkRights(AuthorizationType.Deployer.Edit)
+    correctActor && userService.getCurrentUser.map(_.checkRights(AuthorizationType.Deployer.Edit)).getOrElse(false)
```
clarktsiory, line 341: this is not the way to do this, ideally we pass an `implicit authz: AuthzToken` but this cannot be done without changing an interface inherited from Rudder, see https://github.com/Normation/rudder-plugins/pull/852

fanf (approved):

PR #944, opened 2026-04-03
Fixes #28657: Relay-to-node operation succeeds even if some nodes still depend on relay

skaerg:
https://issues.rudder.io/issues/28657

This PR aims to make the `relay-to-node` command fail if it is run on a relay that is currently managing one or more nodes. Additionally, the change in this PR forbids the `relay-to-node` from being run on the node that has the `root` id.

### DIFF
--- a/scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
+++ b/scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
@@ -132,13 +132,40 @@ class ScaleOutRelayService(
     implicit val qr: QueryContext = cc.toQuery
     for {
       _        <- ScaleOutRelayLoggerPure.debug(s"Start of demoting relay '${nodeId.value}' to node")
+      _        <- if (nodeId == NodeId("root")) {
+                    Inconsistency("Demoting node with id 'root' to simple node is forbidden.").fail
+                  } else ().succeed
       nodeInfo <- nodeFactRepository
                     .get(nodeId)
                     .notOptional(s"Relay with UUID ${nodeId.value} is missing and can not be demoted to node")
+      nodes    <- nodeFactRepository.getAll()
       _        <- if (nodeInfo.rudderSettings.isPolicyServer) {
-                    val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
-                    demoteRelay(nodeInfo, configObjects).unit *>
-                    actionLogger.saveDemoteToNode(cc.modId, cc.actor, nodeInfo.toNodeInfo, cc.message)
+                    val managedNodes = {
+                      Set
+                        .from(
+                          nodes
+                            .filter((_, nodeFact) => nodeFact.rudderSettings.policyServerId == nodeId)
+                            .keySet
+                        )
+                        .excl(nodeId) // node is managed by itself
+                    }
+
+                    if (managedNodes.isEmpty) {
+                      val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
+                      demoteRelay(nodeInfo, configObjects).unit *>
+                      actionLogger.saveDemoteToNode(cc.modId, cc.actor, nodeInfo.toNodeInfo, cc.message)
+                    } else {
+                      val targets = managedNodes.map(n => s"'${n.value}'").mkString(", ")
+                      val errMsg  = s"Relay '${nodeId.value}' cannot be demoted to simple node " +
+                        s"because it is the policy server of the following nodes : \n" +
+                        s"${targets} \n" +
+                        s"These nodes must be assigned to a different policy server before relay '${nodeId.value}' can be demoted."
+
+                      ScaleOutRelayLoggerPure.debug(errMsg) *>
+                      Inconsistency(errMsg).fail
+
+                    }
+
                   } else {
                     ScaleOutRelayLoggerPure.debug(s"Node '${nodeId.value}' is already a simple node, nothing to do.") *>
                     ZIO.unit
--- a/scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/api/ScaleOutRelayApi.scala
+++ b/scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/api/ScaleOutRelayApi.scala
@@ -107,7 +107,7 @@ class ScaleOutRelayApiImpl(
       )
       scaleOutRelayService
         .demoteRelayToNode(NodeId(nodeId))
-        .chainError(s"Error when trying to demote mode $nodeId")
+        .chainError(s"Error when trying to demote node $nodeId")
         .as(nodeId)
         .toLiftResponseOne(params, schema, None)
     }
--- a/scale-out-relay/src/test/resources/scaleoutrelay_api/api_scaleoutrelay.yml
+++ b/scale-out-relay/src/test/resources/scaleoutrelay_api/api_scaleoutrelay.yml
@@ -55,5 +55,35 @@ response:
     {
       "action": "demoteToNode",
       "result": "error",
-      "errorDetails": "Error when trying to demote mode node_void123; cause was: Inconsistency: Relay with UUID node_void123 is missing and can not be demoted to node"
+      "errorDetails": "Error when trying to demote node node_void123; cause was: Inconsistency: Relay with UUID node_void123 is missing and can not be demoted to node"
+    }
+---
+description: Fail to demote root node
+method: POST
+url: /api/latest/scaleoutrelay/demote/root
+headers:
+  - "Content-Type: application/json"
+response:
+  code: 500
+  type: json
+  content: >-
+    {
+      "action": "demoteToNode",
+      "result": "error",
+      "errorDetails": "Error when trying to demote node root; cause was: Inconsistency: Demoting node with id 'root' to simple node is forbidden."
+    }
+---
+description: Fail to demote a relay that is not root to a simple node when the relay is the policy server of one or more nodes
+method: POST
+url: /api/latest/scaleoutrelay/demote/node-dsc
+headers:
+  - "Content-Type: application/json"
+response:
+  code: 500
+  type: json
+  content: >-
+    {
+      "action": "demoteToNode",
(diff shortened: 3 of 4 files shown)


scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -134,10 +134,31 @@ class ScaleOutRelayService(
...
       _        <- if (nodeInfo.rudderSettings.isPolicyServer) {
                     val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
-                    demoteRelay(nodeInfo, configObjects).unit *>
-                    actionLogger.saveDemoteToNode(cc.modId, cc.actor, nodeInfo.toNodeInfo, cc.message)
+                    val targets       = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id).targets
+                    val allNodes      = nodes.mapValues(_.rudderSettings.kind.isPolicyServer)
+                    val managedNodes  = {
+                      RuleTarget
+                        .getNodeIds(targets.toSet, allNodes.toMap, Map())
+                        .filterNot(_.equals(nodeInfo.id))
```
skaerg: i added this line because the API tests for relay promotion/demotion would fail otherwise (the node `node1` on which relay promotion/demotion is tested seems to appear in its own list of managed nodes).

with that said, should it be possible for a relay to have itself as a policy server (except for the rudder server) , i.e. should this test be changed so that `node1` is no longer its own policy server in the API tests ?

scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -134,10 +134,31 @@ class ScaleOutRelayService(
       nodeInfo <- nodeFactRepository
                     .get(nodeId)
                     .notOptional(s"Relay with UUID ${nodeId.value} is missing and can not be demoted to node")
+      nodes    <- nodeFactRepository.getAll()
       _        <- if (nodeInfo.rudderSettings.isPolicyServer) {
                     val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
-                    demoteRelay(nodeInfo, configObjects).unit *>
-                    actionLogger.saveDemoteToNode(cc.modId, cc.actor, nodeInfo.toNodeInfo, cc.message)
+                    val targets       = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id).targets
```
clarktsiory:
```suggestion
                    val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
                    val targets       = configObjects.targets
```

scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
clarktsiory: From the explanation, I understand that "a relay is managed by itself" : in that case it seems fine to just ignore it for the time of "demote" operation

However the `filterNot(_.equals(...))` looks like a `-`/aka `removed(...)`
```suggestion
                        - nodeInfo.id
```

clarktsiory (changes requested):
There are some simplifications that could be applied, but most importantly : would it be possible to add a test case (e.g. in API tests, by creating a test "cluster of nodes", or in another tests file) ?

Testing this case seems important, since it is business logic, and would cover common logic errors (e.g. inverted boolean condition, unexpected cases from the data)

skaerg:
Commit modified

skaerg:
Commit modified

scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -132,13 +132,35 @@ class ScaleOutRelayService(
...
+      nodes    <- nodeFactRepository.getAll()
       _        <- if (nodeInfo.rudderSettings.isPolicyServer) {
-                    val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
-                    demoteRelay(nodeInfo, configObjects).unit *>
-                    actionLogger.saveDemoteToNode(cc.modId, cc.actor, nodeInfo.toNodeInfo, cc.message)
+                    val managedNodes = nodes
+                      .filter((_, nodeFact) => nodeFact.rudderSettings.policyServerId == nodeId)
+                      .map((id, _) => id)
+                      .toSet
+                      - nodeId
```
clarktsiory: directly accessing the map keySet would be the idiomatic and efficient way (`O(1)` instead of `O(n)`)
Also, a code comment explaining the comment that was raised in this PR may always be useful
```suggestion
                      .keySet
                      .removed(nodeId)  // node is managed by itself
```

scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -132,13 +132,35 @@ class ScaleOutRelayService(
...
+                    if (managedNodes.isEmpty) {
+                      val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
+                      demoteRelay(nodeInfo, configObjects).unit *>
+                      actionLogger.saveDemoteToNode(cc.modId, cc.actor, nodeInfo.toNodeInfo, cc.message)
+                    } else {
+                      val targets = managedNodes.map(id => s"'${id}'").reduce((s1, s2) => s"${s1}, ${s2}")
+                      val errMsg  = s"Relay '${nodeId.value}' cannot be demoted to simple node " +
+                        s"because it is the policy server of nodes with the following ids : \n" +
+                        s"${targets} \n" +
+                        s"These nodes must be assigned to a different policy server before Relay '${nodeId.value}' can be demoted."
```
clarktsiory:
```suggestion
                        s"These nodes must be assigned to a different policy server before relay '${nodeId.value}' can be demoted."
```

scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -132,13 +132,35 @@ class ScaleOutRelayService(
...
+                      .map((id, _) => id)
+                      .toSet
+                      - nodeId
+
+                    if (managedNodes.isEmpty) {
+                      val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
+                      demoteRelay(nodeInfo, configObjects).unit *>
+                      actionLogger.saveDemoteToNode(cc.modId, cc.actor, nodeInfo.toNodeInfo, cc.message)
+                    } else {
+                      val targets = managedNodes.map(id => s"'${id}'").reduce((s1, s2) => s"${s1}, ${s2}")
```
clarktsiory: `${id}` would be printed as `NodeId(1234-...)` which is not a proper format for a human readable log. We could just access the value of the node id.
Also, there is the dedicated `mkString` function for the reduce operation:  
```suggestion
                      val targets = managedNodes.map(_.value).mkString(", ")
```

clarktsiory: by the way, we should be careful and explicit when using `reduce`, since it raises an exception when the collection is empty. It should generally be avoided (in favor of `reduceOption`, which makes it catchable at type-level), or at least made explicit in a comment.
Hence, why `mkString` is more pleasing and reassuring 😄

scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -132,13 +132,35 @@ class ScaleOutRelayService(
...
+                      val errMsg  = s"Relay '${nodeId.value}' cannot be demoted to simple node " +
+                        s"because it is the policy server of nodes with the following ids : \n" +
+                        s"${targets} \n" +
+                        s"These nodes must be assigned to a different policy server before Relay '${nodeId.value}' can be demoted."
+
+                      Inconsistency(errMsg).fail *>
+                      ScaleOutRelayLoggerPure.debug(errMsg)
```
clarktsiory: this won't execute the `debug` message : `fail` would short-circuit the `flatMap`/`*>`.
What would do the thing the expected log is to simply switch both lines

clarktsiory (changes requested):
The test addition is nice !

A few remarks that would make it more efficient and idiomatic !

skaerg:
Commit modified

scale-out-relay/src/main/scala/com/normation/plugins/scaleoutrelay/ScaleOutRelayService.scala
```diff
@@ -132,13 +132,35 @@ class ScaleOutRelayService(
...
                     .notOptional(s"Relay with UUID ${nodeId.value} is missing and can not be demoted to node")
+      nodes    <- nodeFactRepository.getAll()
       _        <- if (nodeInfo.rudderSettings.isPolicyServer) {
-                    val configObjects = PolicyServerConfigurationObjects.getConfigurationObject(nodeInfo.id)
-                    demoteRelay(nodeInfo, configObjects).unit *>
-                    actionLogger.saveDemoteToNode(cc.modId, cc.actor, nodeInfo.toNodeInfo, cc.message)
+                    val managedNodes = nodes
+                      .filter((_, nodeFact) => nodeFact.rudderSettings.policyServerId == nodeId)
+                      .keySet
+                      .diff(Set(nodeId)) // node is managed by itself
```
clarktsiory: `diff` allocates an extra "Set", and may iterate (which is what we initially wanted to avoid)

My earlier suggestion for `removed` was not the right one, the right method would be `excl`

skaerg:
Commit modified

skaerg:
Commit modified

clarktsiory (approved):
LGTM, nice addition !

clarktsiory:
OK, merging this PR

PR #946, opened 2026-04-10
Fixes #28491: Accepting a change request that deletes a group results in a dead link in the change request's details page

skaerg:
https://issues.rudder.io/issues/28491

This PR aims to remove links to resources (groups, directives, parameters, rules) that have been deleted in a change request in that change request's page if this change request has been deployed.
This change affects the change request's "Change History" tab as well as the "Diff" tab.

Hence, if the change request has not been deployed yet, the link to the resource will be displayed as before :
image
 
image

However, if the change request has been deployed, the links to the resource will be removed : 
image

image

### DIFF
--- a/change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
+++ b/change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
@@ -1,4 +1,4 @@
-module ChangeRequestChangesForm exposing (Changes, Diff, Directive, DirectiveDiff, GlobalParameter, GlobalParameterDiff, Model, Msg, NodeGroup, NodeGroupDiff, ResourceDiff, Rule, RuleDiff, initModel, update, updateChangeRequestDetails, view)
+module ChangeRequestChangesForm exposing (Changes, Diff, Directive, DirectiveChange, DirectiveDiff, GlobalParameter, GlobalParameterDiff, Model, Msg, NodeGroup, NodeGroupChange, NodeGroupDiff, ParameterChange, ResourceDiff, Rule, RuleChange, RuleDiff, initModel, update, updateChangeRequestDetails, view)
 
 import ErrorMessages exposing (getErrorMessage)
 import Html exposing (Attribute, Html, button, div, h4, li, pre, table, text, ul)
@@ -88,8 +88,7 @@ type alias DiffChange ty =
 
 
 type alias Directive =
-    { id : String
-    , displayName : String
+    { reference : ResourceReference
     , shortDescription : String
     , techniqueName : String
     , techniqueVersion : String
@@ -121,8 +120,7 @@ type alias DirectiveDiff =
 
 
 type alias Rule =
-    { id : String
-    , displayName : String
+    { reference : ResourceReference
     , shortDescription : String
     , longDescription : String
     , targets : TargetList -- List of targets
@@ -150,8 +148,7 @@ type alias RuleDiff =
 
 
 type alias NodeGroup =
-    { id : String
-    , displayName : String
+    { reference : ResourceReference
     , description : String
     , enabled : Bool
     , dynamic : Bool
@@ -179,7 +176,7 @@ type alias NodeGroupDiff =
 
 
 type alias GlobalParameter =
-    { name : String
+    { reference : ResourceReference
     , value : Value
     , description : String
     }
@@ -201,11 +198,27 @@ type ResourceDiff resource modifyDiff
     | Modify modifyDiff
 
 
+type alias DirectiveChange =
+    ResourceDiff Directive DirectiveDiff
+
+
+type alias RuleChange =
+    ResourceDiff Rule RuleDiff
+
+
+type alias NodeGroupChange =
+    ResourceDiff NodeGroup NodeGroupDiff
+
+
+type alias ParameterChange =
+    ResourceDiff GlobalParameter GlobalParameterDiff
+
+
 type alias Changes =
-    { directives : List (ResourceDiff Directive DirectiveDiff)
-    , rules : List (ResourceDiff Rule RuleDiff)
-    , groups : List (ResourceDiff NodeGroup NodeGroupDiff)
-    , parameters : List (ResourceDiff GlobalParameter GlobalParameterDiff)
+    { directives : List DirectiveChange
+    , rules : List RuleChange
+    , groups : List NodeGroupChange
+    , parameters : List ParameterChange
     }
 
 
@@ -305,14 +318,14 @@ update msg model =
 ------------------------------
 
 
-getChangeRequestChanges : Int -> Model -> Cmd Msg
-getChangeRequestChanges crId model =
+getChangeRequestChanges : Int -> String -> Model -> Cmd Msg
+getChangeRequestChanges crId status model =
     request
         { method = "GET"
         , headers = [ header "X-Requested-With" "XMLHttpRequest" ]
         , url = getApiUrl model.contextPath ("changevalidation/workflow/changeRequestChanges/" ++ String.fromInt crId)
         , body = emptyBody
-        , expect = expectJson GetChangeRequestChanges decodeChanges
+        , expect = expectJson GetChangeRequestChanges (decodeChanges status)
         , timeout = Nothing
         , tracker = Nothing
         }
@@ -324,8 +337,8 @@ getChangeRequestChanges crId model =
 ------------------------------
 
 
-decodeDirectiveDiff : Decoder (ResourceDiff Directive DirectiveDiff)
-decodeDirectiveDiff =
+decodeDirectiveDiff : String -> Decoder DirectiveChange
+decodeDirectiveDiff changeRequestStatus =
     let
         decodePolicyMode : Decoder String
         decodePolicyMode =
@@ -339,11 +352,10 @@ decodeDirectiveDiff =
                             fail "Invalid policy mode"
                     )
 
-        decodeDirective : Decoder Directive
-        decodeDirective =
+        decodeDirective : Action -> Decoder Directive
+        decodeDirective action =
             succeed Directive
-                |> requiredAt [ "change", "id" ] string
-                |> requiredAt [ "change", "displayName" ] string
+                |> required "change" (decodeResourceReference "displayName" changeRequestStatus action)
                 |> requiredAt [ "change", "shortDescription" ] string
                 |> requiredAt [ "change", "techniqueName" ] string
                 |> requiredAt [ "change", "techniqueVersion" ] string
@@ -359,10 +371,10 @@ decodeDirectiveDiff =
             (\action ->
                 case action of
                     "create" ->
-                        map Create decodeDirective
+                        map Create (decodeDirective CreateAction)
 
                     "delete" ->
-                        map Delete decodeDirective
+                        map Delete (decodeDirective DeleteAction)
 
                     "modify" ->
                         map Modify
@@ -385,14 +397,13 @@ decodeDirectiveDiff =
             )
 
 
-decodeRuleDiff : Decoder (ResourceDiff Rule RuleDiff)
-decodeRuleDiff =
+decodeRuleDiff : String -> Decoder RuleChange
+decodeRuleDiff changeRequestStatus =
     let
-        decodeRule : Decoder Rule
-        decodeRule =
+        decodeRule : Action -> Decoder Rule
+        decodeRule action =
             succeed Rule
-                |> requiredAt [ "change", "id" ] string
-                |> requiredAt [ "change", "displayName" ] string
+                |> required "change" (decodeResourceReference "displayName" changeRequestStatus action)
                 |> requiredAt [ "change", "shortDescription" ] string
                 |> requiredAt [ "change", "longDescription" ] string
                 |> required "ruleTargetInfo" decodeTargetList
@@ -406,10 +417,10 @@ decodeRuleDiff =
             (\action ->
                 case action of
                     "create" ->
-                        map Create decodeRule
+                        map Create (decodeRule CreateAction)
 
                     "delete" ->
-                        map Delete decodeRule
+                        map Delete (decodeRule DeleteAction)
 
                     "modify" ->
                         map Modify
@@ -430,14 +441,13 @@ decodeRuleDiff =
             )
 
 
-decodeGroupDiff : Decoder (ResourceDiff NodeGroup NodeGroupDiff)
-decodeGroupDiff =
+decodeGroupDiff : String -> Decoder NodeGroupChange
+decodeGroupDiff changeRequestStatus =
     let
-        decodeNodeGroup : Decoder NodeGroup
-        decodeNodeGroup =
+        decodeNodeGroup : Action -> Decoder NodeGroup
+        decodeNodeGroup action =
             succeed NodeGroup
-                |> requiredAt [ "change", "id" ] string
-                |> requiredAt [ "change", "displayName" ] string
+                |> required "change" (decodeResourceReference "displayName" changeRequestStatus action)
                 |> requiredAt [ "change", "description" ] string
                 |> requiredAt [ "change", "enabled" ] bool
                 |> requiredAt [ "change", "dynamic" ] bool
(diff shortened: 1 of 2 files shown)


skaerg:
While the change in this PR prevents a given resource from having a link in the change request that deletes this resource when the change request is deployed, the proposed change in this PR does not prevent links to deleted resources from appearing in other change requests. for example, one could : 
- create a first change request that modifies a given resource
- then, create a second change request that deletes the same resource
- deploy the second change request

if this sequence is reproduced, the first change request will still contain a link to the resource, and the link will be dead since the resource is now deleted.
Preventing this bug from happening will most likely require that the API response be modified in order to account for resources that may have been deleted before a given change request is accepted.

do you think a new issue should be opened in order to fix this problem @clarktsiory ?

skaerg:
Commit modified

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
```diff
@@ -623,28 +623,36 @@ diff model =
...
-                            f2 contextPath r
+                            withDiff contextPath r
 
                 directives =
-                    List.map (defaultOrDiff displayDirective displayDirectiveDiff model.contextPath) data.directives
+                    List.map (displayAction displayDirective displayDirectiveDiff model.contextPath) data.directives
 
                 groups =
-                    List.map (defaultOrDiff displayGroup displayGroupDiff model.contextPath) data.groups
+                    List.map (displayAction displayGroup displayGroupDiff model.contextPath) data.groups
```
clarktsiory: I find this `view` API fragile : it looks like we are doing "business logic" from a decision tree 
1. diff is deletion / on a non-"deployed" state ➡️ display no diff, and link
2. diff is creation ➡️ display no diff with link
3. diff is modification ➡️ display a diff

clarktsiory (commented):
I see the point of that change, but it seems that piece of "reusable diff" module is getting more complex. We would want to re-use it and have no headache around maintaining the "view" part.

So I'm suggesting a more maintainable solution : to reflect that in the "model" that the `diff` function acts upon. It would also allow adding unit tests against these cases (that would be ideal to add in this PR).
Would it be possible to change in this PR ? (we could even discuss and try that in pair-programming)

skaerg:
Commit modified

skaerg:
Commit modified

change-validation/src/main/elm/sources/RudderDataTypes.elm
```diff
@@ -267,6 +273,17 @@ decodeEvent =
...
+        decodeDisplayLink =
+            decodeAction
+                |> andThen
+                    (\action ->
+                        if action == "Delete" && changeRequestDeployed then
+                            succeed NoLink
+
+                        else
+                            succeed WithLink
+                    )
```
clarktsiory: that's business logic within a decoder : it should be tested.

Plus, I'm not convinced that `changeRequestDeployed` should be a boolean : here the logic is `if action == Delete and status == Deployed`, therefore having an "enum" value for the action and a boolean is not very consistent.

Anyway, the model should not be a hidden boolean (NoLink/WithLink) : the architectural problem is about the "model not being the one that we pass to the downward view function"

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
```diff
@@ -729,10 +778,16 @@ displayGroupDiff contextPath group =
...
+displayRule : Bool -> ContextPath -> Rule -> Html Msg
+displayRule link contextPath rule =
     displayResourceDiff "Rule"
-        [ displayField "Rule" (ruleLinkWithId contextPath rule.id rule.displayName)
+        [ displayField "Rule"
+            (if link then
+                ruleLinkWithId contextPath rule.id rule.displayName
+
+             else
+                text rule.displayName
```
clarktsiory: It looks like arguments should be revisited, from what I understand we are not displaying a `Rule`.
 
* `contextPath` is some "context" argument (it would be "implicit" in Scala)
*  so would be the `link` boolean ? But it is not with the right naming convention (`isLink`/`withLink`), but in general boolean overuse is the cause of [antipatterns](https://sporto.github.io/elm-patterns/basic/minimize-boolean.html), in this case it's sad that we have multiple models such as `ResourceLink` and `Bool` for the same thing. But that's not the main concern (see below)
* `rule` is the main object to display, we know the `displayRule` will make use of many of its attributes (but its sad we don't know which ones from the function definition...)

But the signature of `displayRule` is misleading, and therefore this `view` is not "dumb", it does something more than it should, and it forces developers to rely on its implementation rather than simply its signature : 
* having a `type alias Rule = { id: String, displayName: String, category: String ... }` could mean that we display all its fields individually
* but it looks like we are doing more something like 
   
```elm
-- A model for the "View" we are displaying
type ObjectRef =
    RuleLink RuleId DisplayName    -- DisplayName could be String
    | DirectiveLink DirectiveId DisplayName
    | ...
    | DisplayName
  
type alias Rule = { ruleField : ObjectRef, displayName : DisplayName, category : String, ... }


displayRule : ContextPath -> Rule -> Html Msg
displayRule contextPath { ruleField, displayName, category, ... } =
    displayResourceDiff "Rule"
        [ displayField "Rule" (displayObjectRef ruleField)
        , displayStringField "Name" displayName
        ...
```

So, the "link" should have been precomputed into either a "link" with required data (id, name), or just the name

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
```diff
@@ -623,28 +647,41 @@ diff model =
...
+
+                        Delete WithLink r ->
+                            displaySimpleWithLink contextPath r
 
-                        Delete r ->
-                            f1 contextPath r
+                        Delete NoLink r ->
+                            displaySimpleNoLink contextPath r
 
                         Modify r ->
```
clarktsiory, line 648: This is a also doing business logic, but differently from the previous version ! 
There is exactly the same decision tree and the special case for `Delete` in which case we don't have links...

The solution for that is to have it "precomputed in the model", see [this comment](https://github.com/Normation/rudder-plugins/pull/946/changes#r3098813157)

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
```diff
@@ -197,15 +197,31 @@ type alias GlobalParameterDiff =
 
 type ResourceDiff resource modifyDiff
     = Create resource
-    | Delete resource
+    | Delete ResourceLink resource
```
clarktsiory: this is not at the right place : it should be within the `resource` model itself, since we display the link from a field of the `resource`

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
```diff
@@ -504,26 +520,34 @@ decodeParameterDiff =
...
+decodeChanges : String -> Decoder Changes
+decodeChanges changeRequestStatus =
+    let
+        displayDeletedResourceLink =
+            if changeRequestStatus == "Deployed" then
+                NoLink
+
+            else
+                WithLink
+    in
```
clarktsiory: This is closer to the solution we want, we should likely pass the status down the chain : 
* but `changeRequestStatus` should be passed to individual decoding functions (`decodeGroupDiff`, ...)
* then, as explained in [this comment](https://github.com/Normation/rudder-plugins/pull/946/changes#r3098813157), the place where we should do a single `if deployed && delete` is where we decode the model of the object we display, e.g. : 

```elm
decodeGroupDiff : ChangeRequestStatus -> Decoder NodeGroupChange
decodeGroupDiff status =
    field "action" string
        |> andThen
            (\action ->
                case action of
                    "create" ->
                        map Create (decodeNodeGroup status action)

                    "delete" ->
                        map Delete (decodeNodeGroup status action)
                    
                    ...
            )

decodeNodeGroup : ChangeRequestStatus -> Action -> Decoder NodeGroup
decodeNodeGroup status action =
    succeed NodeGroup
        |> required "change" (decodeObjectRef status action)
        |> requiredAt [ "change", "displayName" ] string
        ...


decodeObjectRef : ChangeRequestStatus -> Action -> ObjectRef
decodeObjectRef status action =
   if status == Deployed && action == Delete then
      required "displayName" string
   else
      ... -- decode "id" and "displayName" and create a Link 
```

clarktsiory (changes requested):
The leading line here is : we shouldn't use any `Bool`, except at one single place where the business logic lies.
And it is : 
* `if changeRequestStatus == Deployed && changeRequestAction = Delete`
  * then ~display~ represent the object with a "display string" case
* else
  * represent the object with a "link" field

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
```diff
@@ -406,10 +417,10 @@ decodeRuleDiff =
             (\action ->
                 case action of
                     "create" ->
-                        map Create decodeRule
+                        map Create (decodeRule "create")
 
                     "delete" ->
-                        map Delete decodeRule
+                        map Delete (decodeRule "delete")
```
clarktsiory: It would be the opportunity to pass a dedicated datatype, to avoid using "strings" and mixing-up arguments for the `decodeResourceReference` function

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
```diff
@@ -475,14 +485,22 @@ decodeGroupDiff =
...
     let
-        decodeGlobalParameter : Decoder GlobalParameter
-        decodeGlobalParameter =
+        decodeParamReference : String -> Decoder GlobalParameterReference
+        decodeParamReference action =
+            if changeRequestStatus == "Deployed" && (action == "delete") then
+                succeed ParamPlainText |> required "id" string
+
+            else
+                succeed ParamLink |> required "id" string
```
clarktsiory: I don't see why we need a special case for global parameter : I think it could be merged into the `ResourceReference` type, since some parameters have no name, the case is "link with ID only".

Having something like this would would avoid duplicating the logic :
```elm 
type ResourceReference
    = PlainText String
    | Link ResourceChangeLink
    | LinkNoName String
```

clarktsiory (commented):
LGTM, but I have yet some other suggestions to avoid duplication, and confusing `String`

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
skaerg: i did consider this solution at the start; however, using this datatype would result in a "useless/impossible" match case in the `displayParameter` and `displayResourceLink` functions. here is the current code for the `displayResourceLink` function : 

```elm
displayResourceLink : ContextPath -> ResourceTypeWithId -> ResourceReference -> Html Msg
displayResourceLink contextPath resourceType reference =
    let
        field =
            case reference of
                PlainText displayName ->
                    text displayName

                Link { resourceId, resourceName } ->
                    case resourceType of
                        DirectiveRes ->
                            directiveLinkWithId contextPath resourceId resourceName

                        NodeGroupRes ->
                            groupLinkWithId contextPath resourceId resourceName

                        RuleRes ->
                            ruleLinkWithId contextPath resourceId resourceName

        resourceTypeStr =
            case resourceType of
                DirectiveRes ->
                    "Directive"

                NodeGroupRes ->
                    "Group"

                RuleRes ->
                    "Rule"
    in
    displayField resourceTypeStr field
```

if i were to add the `LinkNoName` sum value type to the `ResourceReference` type, this would create a new `LinkNoName` case in the pattern match, which is an impossible state since this case is not applicable to directives, rules and node groups.

this would have the same effect, only in reverse, on the `displayParameter`  function (since parameters do not have ids, the `Link` case is not applicable to global parameters).

this is the reason why i separated the `ResourceReference` and `GlobalParameterReference` types. i do now think that there might be a way to further factorize these types in a way that should not create the "impossible states" that i mentioned (with a type parameter), so i will try to further factorize the `reference` datatypes in my next commit.

clarktsiory: From my understanding, a "Global parameter" is just a resource whose ID is the same as the name.

And the display is about the same as for resources that have ID, it's just that they don't need a `(Rudder ID: <uuid>)` when displayed : 
Screenshot From 2026-04-20 16-19-11

So, it seems to me that having a "resource without link" is completely a legit case : these "impossible states" are display cases to ignore in the view (in fact, it is an easier model to evolve, for example if some day we assign a different ID than the name for a global parameter).
At least, it is better than having to duplicate and making a special case for every "item that has no ID" (we could imagine that some day we add another resource that has no ID, we would want to avoid duplication of both the decoding logic and the ADTs associated with global parameters)


There may be better ways to represent that, but I fear that it would lead to more complex representation and duplication... If you manage to find one, I would be eager to see what simplification it brings.
But I think minimizing the model and simplifying the special would be the solution for this change

change-validation/src/main/elm/sources/RudderDataTypes.elm
```diff
@@ -250,22 +329,15 @@ decodeEvent =
...
+                        if action == DeleteAction && changeRequestDeployed then
+                            succeed NoLink
 
-                            "modify" ->
-                                succeed "Modify"
-
-                            _ ->
-                                fail "Invalid action"
+                        else
+                            succeed WithLink
```
clarktsiory: Is this not deprecated, because of having the "link/text" already obtained in the `ResourceReference` ?

change-validation/src/main/elm/sources/ChangeRequestChangesForm.elm
clarktsiory: My attempt to simplify the pattern match would be : 
```elm
displayResourceLink : ContextPath -> ResourceTypeWithId -> ResourceReference -> Html Msg
displayResourceLink contextPath resourceType reference =
    let
        field =
            case ( reference, resourceType ) of
                ( PlainText displayName, _ ) ->
                    text displayName

                ( Link { resourceId, resourceName }, DirectiveRes ) ->
                    directiveLinkWithId contextPath resourceId resourceName

                ( Link { resourceId, resourceName }, NodeGroupRes ) ->
                    groupLinkWithId contextPath resourceId resourceName

                ( Link { resourceId, resourceName }, RuleRes ) ->
                    ruleLinkWithId contextPath resourceId resourceName
         
                ( LinkNoName resourceName, ParameterRes ) ->
                    paramLink contextPath resourceName

                -- This is the unrepresentable case
                _ ->
                   ...
...
```

change-validation/src/main/elm/sources/RudderDataTypes.elm
```diff
@@ -216,56 +249,81 @@ compositionDec =
...
+                    (\action ->
+                        if action == DeleteAction && changeRequestDeployed then
+                            map Link
+                                (map2 LinkWithId
+                                    (field "resourceId" string)
+                                    (field "resourceName" string)
+                                )
+
+                        else
+                            map PlainText (field "resourceName" string)
```
clarktsiory: This looks misplaced : it shouldn't be in the `let`, it should be very visible and very similar (even for arguments) to the current :
```elm
decodeResourceReference : String -> Action -> Decoder ResourceReference
decodeResourceReference changeRequestStatus actionType =
    case ( changeRequestStatus, actionType ) of
        ( "Deployed", DeleteAction ) ->
            succeed PlainText |> required "displayName" string

        _ ->
            map Link
                (succeed LinkWithId
                    |> required "id" string
                    |> required "displayName" string
                )
```

change-validation/src/main/elm/sources/RudderDataTypes.elm
```diff
@@ -216,56 +249,77 @@ compositionDec =
...
+        decodeResourceRef action =
+            field "resourceType" string
                 |> andThen
                     (\s ->
                         case s of
                             "directive" ->
-                                succeed DirectiveRes
+                                map2 TypedResourceReference
+                                    (succeed DirectiveRes)
+                                    (decodeDisplayLink action)
```
clarktsiory: could it not be a `succeed TypedResourceReference`, since we mostly use that ?

clarktsiory (commented):
LGTM, it only needs a rebase (then you can assign the "ready for merge" label)

PR #947, opened 2026-04-16
Fixes #28732: Opening user token in main bar is not possible in change logs & archives pages 

clarktsiory:
https://issues.rudder.io/issues/28732


fanf (approved):
These pr are both sad (it was missing just that) and nice (just adding that is enough !)

PR #962, opened 2026-05-20
Fixes #28893: Support JWT in UserInfo endpoint response

clarktsiory:
https://issues.rudder.io/issues/28893

Adding the configuration for support of signing and/or of the UserInfo endpoint, and intercepting the missing handler for a "`application/jwt` content-type response", from the userInfo endpoint.

It is mostly about binding configuration to Nimbus JWT (the library that handles the JOSE spec and allows the decoding JWTs), and creating specific configuration for public/private RSA/EC asymmetric encryption keys (it is included in tests), the spec can be found here: https://datatracker.ietf.org/doc/html/rfc7518.
More signature algorithms are supported with the addition of a `discoverAlgorithms` parameter used for Nimbus configuration. 

The keys configuration could easily be added to JWT/opaque tokens providers too, if we later add support for encryption in them.


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit: add config parsing tests

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -34,19 +34,60 @@
...
+import java.security.PrivateKey
+import java.security.PublicKey
+import java.security.interfaces.ECPrivateKey
+import java.security.interfaces.ECPublicKey
+import java.security.interfaces.RSAPrivateCrtKey
+import java.security.interfaces.RSAPrivateKey
+import java.security.interfaces.RSAPublicKey
+import java.security.spec.ECPoint
+import java.security.spec.ECPublicKeySpec
+import java.security.spec.RSAPublicKeySpec
```
fanf: can you put some stars to limit the boilerplate ? I don't think anyone really carefully look at imports (appart when there is problems)

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -110,6 +151,244 @@ final case class JwtAudience(
...
+
+  val jwkSet: ImmutableJWKSet[SecurityContext] = new ImmutableJWKSet[SecurityContext](
+    JWKSet(keys.map((_, config) => config.jwk).toList.asJava)
+  )
+
+  val jweKeySelector: JWEKeySelector[SecurityContext] = {
+    Monoid.combineAll(for {
+      alg <- jweAlgorithms
+      enc <- jweEncryptionMethods
+    } yield {
```
fanf, line 198: is it really that ? All algo can be mapped with all encryption methods ? It really look strange, I would have expected to have only some valid pairs. 
In any case, it looks like the kind of things we want the user to have leverage on, typically to forbid some kind of pairs/only authorize some. So I fear that will need to be configurable, likely with a default list to make it usable by a normal person who doesn't breath encryption standards.

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -110,6 +151,244 @@ final case class JwtAudience(
...
+      JWEDecryptionKeySelector[SecurityContext](alg, enc, jwkSet): JWEKeySelector[SecurityContext]
+    })
+  }
+
+  val decoder: NimbusJwtDecoder = {
+    NimbusJwtDecoder
+      .withJwkSetUri(jwkSetUri)
+      // default to false for compatibility
+      .pipe(if (discoverJwsAlgorithms.getOrElse(false)) _.discoverJwsAlgorithms() else identity)
```
fanf, line 210: make me wonder if we shouldn't do the same for the autodiscovery of URL based on the well known discory URL - it's an other subject, but it would simplify config in most case.

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -110,6 +151,244 @@ final case class JwtAudience(
...
+              .keyID(kid)
+              .build()
+          )
+      }
+    }
+
+    private[JwtConfig] def loadPublicKey(file: File): IOResult[PublicKey] = {
+      parseKeyFile(file).flatMap {
+        case spki: org.bouncycastle.asn1.x509.SubjectPublicKeyInfo =>
+          converter.getPublicKey(spki).succeed
```
fanf: i'm pretty sure that can fail and throw exception. Everything in bouncycastle does, in unexpected way. It should be in a `IOResult.attempt` like many other things around.

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -110,6 +151,244 @@ final case class JwtAudience(
...
+        keyType: KeyType,
+        keys:    KeyFileConfig
+    ): ValidatedJwk = {
+      keyType match {
+        case KT.RSA =>
+          RSA(keys)(
+            new RSAKey.Builder(pub.asInstanceOf[RSAPublicKey])
+              .privateKey(priv.asInstanceOf[RSAPrivateKey])
+              .keyID(kid)
+              .build()
```
fanf: this can throw exception and needs to be in an `IOResult.attempt`

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -110,6 +151,244 @@ final case class JwtAudience(
...
+          )
+        case KT.EC  =>
+          val ecPub = pub.asInstanceOf[ECPublicKey]
+          val curve = Curve.forECParameterSpec(ecPub.getParams)
+          EC(keys)(
+            new ECKey.Builder(curve, ecPub)
+              .privateKey(priv.asInstanceOf[ECPrivateKey])
+              .keyID(kid)
+              .build()
```
fanf: same

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -110,6 +151,244 @@ final case class JwtAudience(
...
+  final case class PrivateFile(privateFile: File) extends KeyFileConfig(privateFile, None) {
+    def load()(using keyType: KeyType): IOResult[(PublicKey, PrivateKey)] = {
+      ValidatedJwk
+        .loadPrivateKey(privateFile)
+        .map(priv => {
+          keyType match {
+            case KT.RSA =>
+              val privRSA       = priv.asInstanceOf[RSAPrivateCrtKey]
+              val publicKeySpec = new RSAPublicKeySpec(privRSA.getModulus, privRSA.getPublicExponent)
+              val keyFactory    = KeyFactory.getInstance("RSA")
```
fanf: this can throw exceptions and needs to be in an `IOResult.attemp`

fanf (changes requested):
I think you were a bit optimistic about java not throwing things around :) 
Appart that, I think we need to make explicit (with a default sensible list) the list of authorized algo/enc. 

And I will need to find a keycloak to test that 😅

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
clarktsiory: it really is two different things, I couldn't any mention of incompatibility in the specs, and also it is used in an indicative way to "search for keys".
It is nearly the same as in the thing returned by "well-known openid-configuration", which is for example in keycloak: 
```json
  "userinfo_encryption_alg_values_supported": [
    "ECDH-ES+A256KW",
    "ECDH-ES+A192KW",
    "ECDH-ES+A128KW",
    "RSA-OAEP",
    "RSA-OAEP-256",
    "RSA1_5",
    "ECDH-ES"
  ],
  "userinfo_encryption_enc_values_supported": [
    "A256GCM",
    "A192GCM",
    "A128GCM",
    "A128CBC-HS256",
    "A192CBC-HS384",
    "A256CBC-HS512"
  ]
```

clarktsiory: I put some defaults there for recommended ones

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -110,6 +134,282 @@ final case class JwtAudience(
...
+  sealed trait ValidatedJwk {
+    def jwk: JWK
+  }
+
+  // We always expect a private key, and should always have a public and private key
+  sealed trait KeyFileConfig(privateFile: File, publicFile: Option[File]) {
+    def load()(using keyType: KeyType): IOResult[(PublicKey, PrivateKey)]
+  }
+  final case class PublicPrivateFile(public: File, privateFile: File) extends KeyFileConfig(privateFile, Some(public)) {
+    def load()(using KeyType): IOResult[(PublicKey, PrivateKey)] =
```
fanf: why having chosen that approach (behavior in the case class) in place of defining a type class for that ? It's not often that we do that, we prefer to have lean case class without behavior. Is this something special here that I don't see ?

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -110,6 +134,282 @@ final case class JwtAudience(
...
+                val keyFactory = KeyFactory.getInstance("EC")
+                val pub        = keyFactory.generatePublic(pubSpec)
+                (pub, priv)
+            }
+          }
+        })
+    }
+  }
+
+  // curried jwk avoids using full "jwk" in toString, equals: since it has secret data, it should not be output
```
fanf, line 340: nice. I didn't know that bit about second arg list.

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -281,7 +590,15 @@ object RudderRegistrationPropertyCommon {
...
+    A_ENFORCE_ROLES_MAPPING   -> "(default true) if true, restricts roles available by the IdP to the role defined in mapping entitlement. Else the map provides alias for Rudder internal role names.",
+    A_JWT_KEYS                -> "(optional) base property for JWT keys definition. Key ID (\"kid\") are direct sub-properties, and require a key `type`",
+    A_JWT_JWS_DISCOVER        -> "(default false) whether an initial query to the JWKS URI should be made to support other signature algorithms than RS256",
+    A_JWT_JWE_ALG             -> "(optional) algorithms for JWE to attempt to match for JWK defined in configuration and JWKS URL, it could be the ones you known the keys of, or the ones matching the `userinfo_encryption_alg_values_supported` from IdP well-known configuration",
+    A_JWT_JWE_ENC             -> "(optional) encryption methods to attempt to match for JWK defined in configuration and JWKS URL, it could be the ones matching the `userinfo_encryption_enc_values_supported` from IdP well-known configuration",
```
fanf: I think you need to either point to a reference URL, or give common example as of today

fanf: Also, it is not quite clear what that parameter will do. Does that *restrict* the attempt to that list ? What happens when the optional param isn't provided ?

fanf (commented):
Still some question. The design seems better, and the load/validation stepped process for keys looks good. But why putting complicated behavior in case classes, and not a separated type class ? It feels very odd compared to what we do now elsewhere.

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
clarktsiory: I didn't know either, before testing it !

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
clarktsiory: yes it will restrict the attempt to the supported.
Indeed I should mention the defaults as being the recommended ones from the spec !

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
clarktsiory: No, indeed it was an evolution from a first implementation with both mandatory public & private keys.
But a typeclass would certainly be the best choice as of now, I will correct that

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -734,9 +1116,30 @@ class RudderPropertyBasedOAuth2RegistrationDefinition(val registrations: Ref[Lis
...
+          AuthBackendsLoggerPure.logEffect.warn(
+            s"Error when attempting to decode a JWT response, please ensure your JWT keys configuration are right, ${jwtConfig.debugConfiguration}.\nCause: ${ex.getMessage}"
+          )
+        case _                          =>
+          AuthBackendsLoggerPure.logEffect.debug(
+            s"Error when attempting to decode a JWT response: ${err.getMessage}"
+          )
+      }
+    case Right(e)  =>
+      AuthBackendsLoggerPure.logEffect.trace("Successfully decoded JWT with headers : {}", e.getHeaders)
```
fanf: this will need to be changed because of https://github.com/Normation/rudder/pull/7161

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
clarktsiory: I added a type class instead, and put the "public file" VS "private & public files" config in a union type which uses the `Show` type class for the debug string of the configuration

clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM now 🎉

PR #964, opened 2026-05-22
Fixes #28890: Reverse entitlements mapping does not support attribution to all tenants

clarktsiory:
https://issues.rudder.io/issues/28890

Typesafe config :
* will only allow `reverseEntitlements."*"` with quotes because `*` is a special configuration character
* but it will have the key `"\"*\""`

So, we need a special case for the extra quotes in the key: we can only parse it by using `config.getObject`, and we will also need to remap the key to remove the extra quotes.

I also added some documentation regarding that

### DIFF
--- a/auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
+++ b/auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
@@ -367,15 +367,23 @@ object RudderRegistrationPropertyCommon {
     for {
       keySet <- IOResult
                   .attempt(s"Missing key '${path}' for OAUTH2 registration '${base.id}' (${registrationAttributes(key)})")(
-                    config.getConfig(path).entrySet().asScala.map(_.getKey()).toList
+                    config.getObject(path).keySet().asScala.toList
                   )
-                  .catchAll(_ =>
-                    List().succeed
-                  ) // in that case, we suppose that the key is just missing so we return a default empty value for mapping
+                  // in that case, we suppose that the key is just missing so we return a default empty value for mapping
+                  .catchAll(_ => List().succeed)
       values <- ZIO.foreach(keySet) { key =>
                   val wholeKey = path + "." + key
                   IOResult.attempt(s"Error when reading role entitlement mapping '${wholeKey}'") {
-                    (key, config.getString(wholeKey))
+                    val (k, v) = {
+                      // see https://issues.rudder.io/issues/28890:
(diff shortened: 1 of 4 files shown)


auth-backends/src/test/scala/com/normation/plugins/authbackends/TestReadOidcConfig.scala
```diff
@@ -88,9 +88,11 @@ class TestReadOidcConfig extends Specification {
         "CN=AAAA-BBBBB,OU=Groups,OU=_IT,OU=BB-DD,OU=UUU-XXXX-YY,DC=ee,DC=if,DC=ttttt,DC=uuu" -> "administrator"
       )) and (
         regs("someidp").tenants.mapping === Map(
+          "rudder_all"                                                                         -> "*",
           "rudder_TA"                                                                          -> "TA",
           "rudder_TB"                                                                          -> "TB_OVERRIDDEN",
-          "CN=AAAA-BBBBB,OU=Groups,OU=_IT,OU=BB-DD,OU=UUU-XXXX-YY,DC=ee,DC=if,DC=ttttt,DC=uuu" -> "TA"
+          "CN=AAAA-BBBBB,OU=Groups,OU=_IT,OU=BB-DD,OU=UUU-XXXX-YY,DC=ee,DC=if,DC=ttttt,DC=uuu" -> "TA",
+          "CN=AAAA-BBBBB,OU=Groups,OU=_IT,OU=BB-DD,OU=Admin"                                   -> "*"
```
clarktsiory: this ensures that we obtain `"*"`, it was `"\"*\""` before the correction

auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -367,15 +367,24 @@ object RudderRegistrationPropertyCommon {
...
                     List().succeed
                   ) // in that case, we suppose that the key is just missing so we return a default empty value for mapping
       values <- ZIO.foreach(keySet) { key =>
                   val wholeKey = path + "." + key
                   IOResult.attempt(s"Error when reading role entitlement mapping '${wholeKey}'") {
-                    (key, config.getString(wholeKey))
+                    val (k, v) = {
+                      // see https://issues.rudder.io/issues/28890:
+                      // tenants can use special config for 'reverseEntitlements."*"' key, but typesafe would keep the double quotes
+                      if (key == "\"*\"") {
```
fanf: I think you need to add "-" in the same case, to be able to manage right control on groups. And just because else, it's inconsistent.

fanf (changes requested):
You need to take care of the other special char (`-`) to  be consistent.

PR #966, opened 2026-05-29
Fixes #28936: Reverse entitlements doc needs line breaks for bullet points

clarktsiory:
https://issues.rudder.io/issues/28936

A link was also broken


fanf (approved):
ok

PR #967, opened 2026-05-29
Fixes #28915: Add example doc for keycloak OIDC identity provider

clarktsiory:
https://issues.rudder.io/issues/28915

Replaced the "google" example that was put in parallel to Okta in the configuration file by "keycloak" (which is known to work).
I even put the Keycloak example configuration first, before the one for Okta, since it could be easier to get started with


fanf (approved):
👌

PR #971, opened 2026-06-09
Fixes #29026: Remove auth backends from rudder-plugins

VinceMacBuche:
https://issues.rudder.io/issues/29026


fanf (approved):
And so be it

PR #973, opened 2026-06-11
Fixes #28784: Header buttons are wrapped when change request name is too long on medium screen

skaerg:
https://issues.rudder.io/issues/28784

### before : 
image

---

### after : 

image

image


skaerg:
i used css classes that looked appropriate in bootstrap-grid.css, but is there a simpler/more "standard" way of achieving this result ?

fanf (approved):
This look appropriate enough ;)

PR #974, opened 2026-06-25
Fixes #29113: Security improvements for OIDC config

clarktsiory:
https://issues.rudder.io/issues/29113

* follow the documented "restrict mapping as true" by default as in the doc
* add tests against cached API opaque token cache to take API account status into consideration   

---
Upmerge will be complicated :sweat_smile: (ApiAccount model has changed in 9.1, and we need a special PR in 9.2)


fanf (approved):
LGTM ! Good luck with the upmerge

clarktsiory:
OK, merging this PR

PR #978, opened 2026-06-30
Fixes #29131: Improve configuration parsing

clarktsiory:
https://issues.rudder.io/issues/29131

* Add an error log upon unknown boolean value which had some default
```
ERROR auth-backends - Unknown configuration value for rudder.auth.oauth2.provider.keycloak.roles.enabled: expected true or false, but got 1. Defaulting to false
```
* Fix the scope parsing and add generic tests with zio-tests


fanf (approved):

clarktsiory:
OK, merging this PR

PR #980, opened 2026-07-02
Fixes #29173: There is a space above change request header buttons since #28784

clarktsiory:
https://issues.rudder.io/issues/29173

Using flex prevents the misalignment, and makes the header take full space (and flex-nowrap prevents surprising wraps).

It now looks like:
image


clarktsiory:
PR updated with a new commit

skaerg (approved):

clarktsiory:
OK, squash merging this PR

PR #985, opened 2026-07-12
Fixes #29240: self-validation parameter is ignored

fanf:
https://issues.rudder.io/issues/29240

Based on https://github.com/Normation/rudder/pull/7290
There is two commit, the first one is the actual implementation of the check - nothing complicated, we just use the check about author/authenticated user introduced in the other PR. See  `checkSelfAction`
The second introduce unit testing to avoid regression on that part. 

Also propagate deletion of `isEditable`. Let me know if it was used somewhere in the UI I didn't find.

### DIFF
--- a/change-validation/src/test/scala/com/normation/plugins/changevalidation/SelfValidationWorkflowTest.scala
+++ b/change-validation/src/test/scala/com/normation/plugins/changevalidation/SelfValidationWorkflowTest.scala
@@ -1,0 +1,284 @@
+/*
+ *************************************************************************************
+ * Copyright 2026 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package com.normation.plugins.changevalidation
+
+import com.normation.errors.*
+import com.normation.plugins.AlwaysEnabledPluginStatus
+import com.normation.plugins.changevalidation.TwoValidationStepsWorkflowServiceImpl.*
+import com.normation.plugins.changevalidation.api.ChangeRequestApiImpl
+import com.normation.plugins.changevalidation.api.ChangeValidationSecurityError
+import com.normation.rudder.AuthorizationType
+import com.normation.rudder.Rights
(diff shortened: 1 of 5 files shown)


fanf:
PR updated with a new commit

change-validation/src/test/scala/com/normation/plugins/changevalidation/SelfValidationWorkflowTest.scala
```diff
@@ -0,0 +1,308 @@
...
+    // Matrix A: findNextSteps at "Pending validation" — the "validate" transition to
+    // "Pending deployment" is gated by self-validation when the user is the author.
+    // ---------------------------------------------------------------------------------------
+    suite("A - findNextSteps: self-validation gates the validate transition")(
+      test("author CAN validate when self-validation is enabled") {
+        implicit val user: AuthenticatedUser = validatorAndDeployer
+        val targets = nextTargets(mkService(enabled, enabled), Validation.id, ChangeRequestAuthorship.Author)
+        assertTrue(targets.contains(Deployment.id))
+      },
+      test("author CANNOT validate their own change when self-validation is disabled (the regressed case)") {
```
clarktsiory: If we mention the regression I think we also want to mention the specific bug-tracker issue (instead of relying on git blame/history)

change-validation/src/test/scala/com/normation/plugins/changevalidation/SelfValidationWorkflowTest.scala
```diff
@@ -0,0 +1,308 @@
...
+        implicit val user: AuthenticatedUser = validatorAndDeployer
+        val targets = nextTargets(mkService(disabled, enabled), Validation.id, ChangeRequestAuthorship.Author)
+        assertTrue(!targets.contains(Deployment.id))
+      },
+      test("a non-author CAN validate even when self-validation is disabled") {
+        implicit val user: AuthenticatedUser = validatorAndDeployer
```
clarktsiory:
```scala
        implicit val user: AuthenticatedUser = validatorAndDeployer
```
seems to be repeated in all tests, it could just be a field-level implicit in the test class ?

change-validation/src/test/scala/com/normation/plugins/changevalidation/SelfValidationWorkflowTest.scala
```diff
@@ -0,0 +1,308 @@
...
+        implicit val user: AuthenticatedUser = userWith(Set(AuthorizationType.Deployer.Edit))
+        val targets = nextTargets(mkService(enabled, enabled), Validation.id, ChangeRequestAuthorship.NotAuthor)
+        assertTrue(!targets.contains(Deployment.id))
+      },
+      test("no deploy transition without Deployer.Edit, even for a non-author with self-deployment enabled") {
+        implicit val user: AuthenticatedUser = userWith(Set(AuthorizationType.Validator.Edit))
+        val targets = nextTargets(mkService(enabled, enabled), Deployment.id, ChangeRequestAuthorship.NotAuthor)
+        assertTrue(!targets.contains(Deployed.id))
+      }
+    ),
```
clarktsiory, line 240: IDK to what more extent zio-test generators would help for matrix tests, but here it could be something like
```scala
checkAll(Gen.elements(Validation.id, Deployment.id)) { step =>
  val targets = nextTargets(..., step, ...)
  assertTrue(!targets.contains(Deployed.id) && !targets.contains(Deployment.id))
}
```

change-validation/src/test/scala/com/normation/plugins/changevalidation/SelfValidationWorkflowTest.scala
```diff
@@ -0,0 +1,308 @@
...
+      test("the change request owner is the acting user -> Author") {
+        val cr = RollbackChangeRequest(ChangeRequestId(1), None, ChangeRequestInfo("cr", "desc"), owner = "alice")
+        assertTrue(ChangeRequestAuthorship.of(cr, userWith(Set.empty, userLogin = "alice")) == ChangeRequestAuthorship.Author)
+      },
+      test("the change request owner is someone else -> NotAuthor") {
+        val cr = RollbackChangeRequest(ChangeRequestId(1), None, ChangeRequestInfo("cr", "desc"), owner = "alice")
+        assertTrue(ChangeRequestAuthorship.of(cr, userWith(Set.empty, userLogin = "bob")) == ChangeRequestAuthorship.NotAuthor)
+      }
+    )
+  )
```
clarktsiory: I find it weird that it's within the list of matrix tests, since the constructor and test subject is defined in https://github.com/Normation/rudder/pull/7290, it would ideally belong and be maintained in Rudder, the place where it's defined

clarktsiory (commented):
LGTM overall, that fix is really a nice one.

A few considerations about matrix tests

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #986, opened 2026-07-12
Fixes #29243: Tenant configuration for auto-provided users are not taken into acount when override is enforced

fanf:
https://issues.rudder.io/issues/29243

Some code for tenant default predated the changes in tenant management with OIDC in 8.3. That lead to a too large default with some cases of configuration options.
Now, we have everything in `RudderUserDetails`, so just use that.

### DIFF
--- a/auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
+++ b/auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
@@ -199,7 +199,7 @@ final case class RudderClientRegistration(
 
   // avoid that in logs etc, use only for interactive debugging sessions
   def toDebugStringWithSecret: String =
-    s"""{${registration.toString}}, '${infoMsg}', roles: ${roles.toString}, user provisioning enabled: ${provisioning}"""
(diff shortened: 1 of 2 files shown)


auth-backends/src/main/scala/com/normation/plugins/authbackends/Oauth2Authentication.scala
```diff
@@ -199,7 +199,7 @@ final case class RudderClientRegistration(
 
   // avoid that in logs etc, use only for interactive debugging sessions
   def toDebugStringWithSecret: String =
-    s"""{${registration.toString}}, '${infoMsg}', roles: ${roles.toString}, user provisioning enabled: ${provisioning}"""
+    s"""{${registration.toString}}, '${infoMsg}', roles: ${roles.toString}, tenants: ${tenants.toString}, user provisioning enabled: ${provisioning}"""
```
fanf, line 202: missing trace log for tenants

clarktsiory (commented):
LGTM, oups...

clarktsiory (approved):

fanf:
OK, merging this PR

PR #987, opened 2026-07-12
Fixes #29245: Missing tenant checking for node external reports

fanf:
https://issues.rudder.io/issues/29245

Several things to hardened node external reports: 
- don't give bad advice: use the plugin default var path in doc/examples, 
- resolve file name server side from node Id / node hostname and plugin config
- user needs `Node.Read` right to see the node reports, 
- paths are sanitized


clarktsiory (approved):
LGTM !

fanf:
OK, merging this PR

PR #1007, opened 2026-08-06
Fixes #29470: webapp non-root: fix broken postinst in plugins

m4rtinh4rt:
https://issues.rudder.io/issues/29470


amousset:
why did it fail?

m4rtinh4rt:
The issue comes from the schema file `/opt/rudder/share/plugins/datasources/datasources-schema.sql`, which is only accessible by root.

PR #1008, opened 2026-08-10
Fixes #29479: SSO provisioned users see 0 node by default

clarktsiory:
https://issues.rudder.io/issues/29479

When tenants are not supposed to be configured, the default logic was "use the tenant `-`" every time (both for JWT and OAuth2/OIDC user login).

That logic branch is not the expected behavior, the fallback when tenants are not enabled should be the retro-compatible behavior of provisioning user with "all tenants", upon every login/JWT authentication.

**Change:**
* fix the branch of logic where `tenants.enable=false`, to use `*` tenants
* fix JWT to use the default based on `tenants.enable` configuration too
* add tests of the `tenants.enable=false` case

### DIFF
--- a/auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
+++ b/auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
@@ -809,9 +809,29 @@ object RudderTokenMapping {
   }
 
   /*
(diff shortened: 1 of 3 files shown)


clarktsiory:
PR updated with a new commit

fanf (approved):
Lgtm, looks like the smallest change for that pb

clarktsiory:
OK, squash merging this PR

PR #1018, opened 2026-08-19
Fixes #29548: Manual tenant config is ignored on OIDC provisioned users

clarktsiory:
https://issues.rudder.io/issues/29548

For now that the parsed user tenant from base is not an `Option`, we need to get the parsed user from file:
* if it's there and has a tenant, fallback must be that one
* if not, fallback is same as for compat: `*`

### DIFF
--- a/auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
+++ b/auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
@@ -809,9 +809,11 @@ object RudderTokenMapping {
   }
 
   /*
-   * Enforces that when tenants are disabled, give access by default to "all tenants".
+   * Enforces that when tenants are disabled, give access by default to "all tenants", if no tenants are found in token.
    *
    * To be used when there are no known default tenants.
+   * E.g. for JWT, if tenants are not provisioned, keep the same behavior as an API account with no tenant attribute,
+   * that is '*'
    */
   def getTenants(
       reg:             RudderOAuth2Registration & RegistrationWithRoles,
@@ -895,8 +897,8 @@ object RudderTokenMapping {
       tenants
 
     } else {
-      // compatibility: disabled tenants configuration means access to all tenants
-      val fallback = NodeSecurityContext.All
+      // disabled tenants configuration means tenant should be resolved as the default
+      val fallback = default
       AuthBackendsLogger.debug(
         s"${protocolName} configuration is not configured to use token provided tenants, falling back to '${fallback.serialize}'"
       )
@@ -991,7 +993,7 @@ trait RudderUserServerMapping[R <: OAuth2UserRequest, U <: OAuth2User, T <: Rudd
     val optReg = registrationRepository.registrations.get(userRequest.getClientRegistration.getRegistrationId)
 
     // check that we know that user in our DB, else if "provisioning" is enabled, create it
-    val rudderUser = {
+    val rudderUser   = {
       try {
         rudderUserDetailsService.loadUserByUsername(user.getName)
       } catch {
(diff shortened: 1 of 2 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit: format

auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
```diff
@@ -1015,16 +1017,58 @@ trait RudderUserServerMapping[R <: OAuth2UserRequest, U <: OAuth2User, T <: Rudd
...
     }
+    // tenants compatibility because of non-optional tenant model in RudderUserDetail: we may need user tenant from file
+    def resolvedUser = if (optReg.map(_.tenants.enabled).getOrElse(false)) {
+      // ok because tenants in file are ignored/merged, see `RudderTokenMapping.getTenants` for the tenant provisioning
+      rudderUser
+    } else {
+      // tenants in base, if present, are the only source of truth since tenant are not provisioned:
+      // - default tenant, if absent (even from file), is '*'
+      // - invalid tenants should be resolved as '-'
+      rudderUserDetailsService.authConfigProvider.authConfig match {
```
fanf: I think that is the same as 
```
rudderUserDetailsService.authConfigProvider.getUserByName(rudderUser.getUsername) match
     case Left(_)    => // user not in XML file, use default
                  AuthBackendsLogger.debug(
              s"User '${user.getName}' is not in file or has no 'tenants' defined, falling back to ${defaultTenant.serialize}"
            )
            rudderUser.copy(nodePerms = defaultTenant)
                case Right(u) => // user in XML file, use its tenants
                  AuthBackendsLogger.debug(
                    s"User '${user.getName}' has resolved tenants from file: ${value.serialize}"
                  )
                  // provided user already has the right tenant, no need for copy
                  rudderUser
              }
```

auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
```diff
@@ -1015,16 +1017,58 @@ trait RudderUserServerMapping[R <: OAuth2UserRequest, U <: OAuth2User, T <: Rudd
...
-    buildUser(optReg, userRequest, user, roleApiMapping, rudderUser, newUserDetails)
+        // unreachable for OIDC/OAuth2
+        case _:         SingleUserList    => rudderUser
+      }
+    }
+
+    buildUser(optReg, userRequest, user, roleApiMapping, resolvedUser, newUserDetails)
   }
 
+  // resolved user can have default fields that we don't want to compute
```
fanf, line 1046: Is this right? It seems that if `resolvedUser` is a val and so computed only once, then whatever it computes is needed to have the correct tenants. 
Plus having a val and a by value param ensure that all that computation is done only once, and so we don't have a strange case where some tenant/session value change (typically, because of a race on the file writing or whatever). 
I think I would prefer to go with the simple and very debugable way here

fanf (commented):
I think the logic is ok, but I wonder if there isn't more complexity than needed. 

We will need to be careful when/if we change all that to `Option`, we will need to rebuild/recheck all that for what should be the default values.

auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
clarktsiory: okay, it looked to me like an optimization but it adds that inconvenience of debuggability...

auth-backends/src/main/scala/bootstrap/rudder/plugin/AuthBackendsConf.scala
clarktsiory: indeed, the invalid case was swallowed...

clarktsiory:
PR updated with a new commit: simplifications

fanf (approved):
LGTM

clarktsiory:
OK, squash merging this PR

PR #1019, opened 2026-08-20
Fixes #29590: Handle trivy checks in public plugins

clarktsiory:
https://issues.rudder.io/issues/29590

### DIFF
--- a/auth-backends/pom-template.xml
+++ b/auth-backends/pom-template.xml
@@ -81,7 +81,7 @@
     <dependency>
       <groupId>com.nimbusds</groupId>
       <artifactId>nimbus-jose-jwt</artifactId>
-      <version>9.45</version>
+      <version>10.5</version>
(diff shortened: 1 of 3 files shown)


auth-backends/pom-template.xml
```diff
@@ -81,7 +81,7 @@
     <dependency>
       <groupId>com.nimbusds</groupId>
       <artifactId>nimbus-jose-jwt</artifactId>
-      <version>9.45</version>
+      <version>10.5</version>
```
clarktsiory, line 84: This one is the exact same version as in the 9.1 branch (hence it's simply a backport): in 9.0 we have the same exact Spring and Spring Security versions as in 9.1
See https://github.com/Normation/rudder-plugins/pull/886#discussion_r2489065353

amousset (approved):

PR #1021, opened 2026-08-24
Fixes #29607: Allow to include vendor js in plugins build 

clarktsiory:
https://issues.rudder.io/issues/29607

Same as in Rudder: have a vendor_js build task with some list of dependencies, which are lazy-loaded from plugin's new `dependencies.mjs` file. It should export a `vendor_js` list.

Dependencies are bundled into `resources/toserve` (without the plugin namespace), see https://github.com/Normation/rudder-plugins-private/pull/1471 for an example usage.

### DIFF
--- a/plugins-common/gulpfile.mjs
+++ b/plugins-common/gulpfile.mjs
@@ -30,7 +30,18 @@ var grep = function(regex) {
 
 const __dirname = path.resolve();
 
+let VENDOR_JS = [];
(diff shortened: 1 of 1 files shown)


amousset (approved):

clarktsiory:
PR rebased

clarktsiory:
OK, merging this PR

